Issue #10, opened 2024-02-13
Append Events with an expected version


watfordsuzy:
I'm looking through the [API Docs while playing with Emmett](https://event-driven-io.github.io/emmett/api-docs.html#event-store) and am wondering about the elision of `expectedVersion` on the `EventStore` interface:
https://github.com/event-driven-io/emmett/blob/9650af6625f2b2065dcad1db2c96aa6495c6afa2/packages/emmett/src/eventStore/eventStore.ts#L3-L20

It is my assumption this interface would instead look something like:
```ts
// #region event-store
export interface EventStore {
  aggregateStream<Entity, E extends Event, NextExpectedVersion = bigint>(
    streamName: string,
    options: {
      evolve: (currentState: Entity, event: E) => Entity;
      getInitialState: () => Entity;
      startingVersion?: NextExpectedVersion | undefined,
    },
  ): Promise<{ entity: Entity | null, nextExpectedVersion: NextExpectedVersion }>;

  readStream<E extends Event, NextExpectedVersion = bigint>(
    streamName: string,
    startingVersion?: NextExpectedVersion | undefined
  ): Promise<E[]>;

  appendToStream<E extends Event, NextExpectedVersion = bigint>(
    streamId: string,
    expectedVersion?: NextExpectedVersion | undefined,
    ...events: E[]
  ): Promise<NextExpectedVersion>;
}
// #endregion event-store
```
Is this a conscious design choice to represent a simpler model of event sourcing?

oskardudycz:
@watfordsuzy, let me quote my response on [your PR](https://github.com/event-driven-io/emmett/pull/11#pullrequestreview-1879616164):

> @watfordsuzy, thank you a lot for the PR and for the issue. Indeed, the lack of optimistic concurrency is a big thing. I didn't add that initially without the intention to hide that part or ignore it. The first release was focused on the business logic; now it's time to do the application layer correctly.
>
> I'm planning to do it more or less like in the solution in my self-paced kit: https://github.com/oskardudycz/EventSourcing.NodeJS/blob/2d417bd2e650c16b4af68b6a752bd684ce0a8fc0/workshops/introduction_to_event_sourcing/src/solved/07_optimistic_concurrency_eventstoredb/tools/eventStore.ts#L99
>
> I'm going to pull your changes in and restructure them a bit. Thank you a lot for the contribution and double-checking if I didn't miss it!

I'm going to work on that today and tag you also in the PR that will fully introduce that 🙂

[...]

PR #11, opened 2024-02-13
Add expected version to EventStore interface #10

watfordsuzy:
This PR is merely to discuss issue #10 and I don't feel confident enough in the foundational API shapes to suggest otherwise.

Trying to adapt Emmett to some problems I found myself smuggling expected versions within event metadata rather than the API itself, while also having the next expected version returned through the API.

### DIFF
--- a/packages/emmett/src/commandHandling/handleCommand.ts
+++ b/packages/emmett/src/commandHandling/handleCommand.ts
@@ -3,7 +3,7 @@ import type { Event } from '../typing';
 
 // #region command-handler
 export const CommandHandler =
-  <State, StreamEvent extends Event>(
+  <State, StreamEvent extends Event, NextExpectedVersion = bigint>(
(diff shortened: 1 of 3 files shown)


oskardudycz (approved):
@watfordsuzy, thank you a lot for the PR and for the issue. Indeed, the lack of optimistic concurrency is a big thing. I didn't add that initially without the intention to hide that part or ignore it. The first release was focused on the business logic; now it's time to do the application layer correctly. 

I'm planning to do it more or less like in the solution in my self-paced kit: https://github.com/oskardudycz/EventSourcing.NodeJS/blob/2d417bd2e650c16b4af68b6a752bd684ce0a8fc0/workshops/introduction_to_event_sourcing/src/solved/07_optimistic_concurrency_eventstoredb/tools/eventStore.ts#L99

I'm going to pull your changes in and restructure them a bit. Thank you a lot for the contribution and double-checking if I didn't miss it!

PR #14, opened 2024-02-14
Refactor/use node native test runners

### LINKED ISSUE #9: Use Node's Native Test Runner instead of Jest
Hey there, Oskar - I really appreciate the effort of creating a Node version of Marten, which is something I've been missing in the JS ecosystem. As a fellow TS programmer and Event-sourcing enthusiast, I'd like to contribute to this tool - and my first suggestion is to use [Node's native test runner](https://nodejs.org/api/test.html) to run its tests. It is way faster than Jest, and it's one less dependency in the repo, which is always great.

I've created a [repo](https://github.com/thiagomini/typescript-node-native-tests) showing how easy it is to set it up with Typescript, and you'll see that it's API is basically the same as Jest or other famous test runners (like Vitest). The idea is to improve the DevEx for the contributors of this package by leveraging native tools. 

I can create a PR to configure the repo to use it - just let me know if you agree!

thiagomini:
Closes #9

Removes Jest in favor of node's [native test runner](https://nodejs.org/api/test.html). The main difference here is that we'll need to use node's internal [assert](https://nodejs.org/api/assert.html) library instead of Jest's "expect" API, but that's a reasonable trade-off IMHO.

PS: I'll manage the merge conflicts soon


package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@event-driven-io/core",
   "version": "0.1.7",
   "description": "Emmett - Event Sourcing development made simple",
+  "engines": {
+    "node": ">=21.6.2"
```
thiagomini, line 6: Hey, @oskardudycz - this is something that needs you attention. In order to use Node's test runner along with its latest features, we need to use the "Current" version of [Node](https://nodejs.org/en). Let me know if that's okay

package.json
```diff
@@ -39,27 +42,24 @@
...
     "@types/node": "20.11.17",
     "@types/uuid": "9.0.8",
     "@typescript-eslint/eslint-plugin": "6.21.0",
     "@typescript-eslint/parser": "6.21.0",
     "eslint": "8.56.0",
     "eslint-config-prettier": "9.1.0",
     "eslint-plugin-prettier": "5.1.3",
-    "jest": "29.7.0",
-    "npm-run-all": "4.1.5",
+    "npm-run-all2": "6.1.2",
```
thiagomini, line 56: I decided to remove the [npm-run-all](https://www.npmjs.com/package/npm-run-all) library since it's not being maintained for 5 years now - the `npm-run-all2` one is a fork with dependabots updates (and some QOF enhancements as well)

thiagomini:
Hey @oskardudycz , this is ready for review, let me know if you have any more thoughts on that!

packages/emmett-esdb/src/testing/example.spec.ts
thiagomini, line 1: Notice the file's extension (.spec) - this is a more common pattern in the JS/TS world, so I decided to use it here

package.json
oskardudycz: @thiagomini my main concern is not to lost or be able to catch other 21.x specifics for packages so as not to accidentally make it incompatible with the LTS version. Afaik some runtimes like AWS Lambda doesn't support 21.x yet and I’d like to have it compatible.

I know that for now it’ll be used only in tests, so in theory that’s fine. 

Thoughts?

package.json
oskardudycz: Fine by me 👌

packages/emmett-esdb/src/testing/example.spec.ts
oskardudycz: Saw that, I don’t mind that 🙂

package.json
thiagomini: I understand your concern, @oskardudycz - There's a way to use LTS here, but we'd have to change the `test` script in the root package.json. I noticed you changed it to not run the `test` script of each workspace, but rather to execute the tests directly - this doesn't work with LTS version of node because the test runner doesn't recognize glob patterns yet (see https://github.com/nodejs/node/issues/50658).

An alternative would be to roll back the root package.json so it calls each workspace "test" script

oskardudycz: I did that to make possible running the Jest plugin for tests and enable debugging. I don’t mind changing that if it makes it work. Did you happen to check tests debugging capabilities?

thiagomini: I see, Oskar! Yea, it's possible to debug a test, we just need to attach a debugger - or if you are using vscode, open a "Javascript debugger terminal" and execute the test command, it will automatically attach a debugger and stop on breakpoints

oskardudycz:
@thiagomini I fixed failing CI by:
- updating GitHub Actions config,
- adjusting nvm settings,
- added pattern to also run tests in the docs folder,
- Refactored the naming pattern to not use dash but dot (e.g. use e2e.spec.ts instead of e2e-spec.ts)

I also tried configuring tests debugging in VSCode, but it didn't work as the breakpoint was not caught. Could you check that? That's the last blocker for me atm.

thiagomini:
Hey @oskardudycz , the configuraiton "Debug Current File" already works for test files. You can just open the test file you want to debug, run that configuration, and Node will stop at the breakpoint. We don't need special configurations since these are interpreted by Node directly - or is it something else you'd like to add?

oskardudycz:
@thiagomini in my case tests were run, but the vscode didn't stop on breakpoint. I was running that on Windows, will try tomorrow on the Linux box.

oskardudycz:
@thiagomini, I checked on my Linux box, and unfortunately, it's the same result. Is it working on your side? Might be that some ts config or workspace configuration is messing with it, but I'm not sure of the reason.

packages/emmett-esdb/src/testing/example.e2e.spec.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+describe('Example e2e Test', () => {
+  it('should pass', () => {
+    assert.equal(1, 1);
+  });
+});
```
marcingolenia, line 9: Hi guys. Sorry for jumping in, isn't this a leftover?

packages/emmett-esdb/src/testing/example.int.spec.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+describe('Example Integration Test', () => {
+  it('should pass', () => {
+    assert.equal(1, 1);
+  });
+});
```
marcingolenia, line 9: leftover?

packages/emmett-esdb/src/testing/example.unit.spec.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+describe('Example Test', () => {
+  it('should pass', () => {
+    assert.equal(1, 1);
+  });
+});
```
marcingolenia, line 9: leftover?

packages/emmett/src/testing/example.int.spec.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+describe('Example Integration Test', () => {
+  it('should pass', () => {
+    assert.equal(1, 1);
+  });
+});
```
marcingolenia, line 9: leftover?

packages/emmett/src/testing/example.e2e.spec.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+describe('Example e2e Test', () => {
+  it('should pass', () => {
+    assert.equal(1, 1);
+  });
+});
```
marcingolenia, line 9: leftover?

packages/emmett-expressjs/src/testing/example.int.spec.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+describe('Example Integration Test', () => {
+  it('should pass', () => {
+    assert.equal(1, 1);
+  });
+});
```
marcingolenia, line 9: leftover?

packages/emmett-expressjs/src/testing/example.e2e.spec.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+describe('Example e2e Test', () => {
+  it('should pass', () => {
+    assert.equal(1, 1);
+  });
+});
```
marcingolenia, line 9: leftover?

thiagomini:
Hey, @oskardudycz , sorry about the late reply - yes, It's working for me when I go to a specific test file and press F5 to run "Debug Current File" configuration, as you can see in the image below. However, I noticed that running `npm test` in the debugger terminal isn't attaching the debugger 🤔 I'll investigate it further.

image

packages/emmett-esdb/src/testing/example.e2e.spec.ts
thiagomini: Yea, we can remove them, they were just examples of how to create tests using the native test runner

.github/workflows/build_and_test.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         # selected operation systems to run CI
         os: [ubuntu-latest] #, windows-latest, macos-latest]
         # selected node version to run CI
-        node-version: [20.10.x]
+        node-version: [20.11.1]
```
thiagomini, line 22: I decided to revert to LTS version. One of the issues with that version was that it didn't work with globstar patterns (see https://github.com/nodejs/node/issues/44023). However, I found a solution by using the [node-glob](https://github.com/isaacs/node-glob) package, which is lightweight and constantly updated.

.vscode/launch.json
```diff
@@ -2,29 +2,46 @@
...
+        // Node.js internal core modules
+        "<node_internals>/**",
+
+        // Ignore all dependencies (optional)
+        "${workspaceFolder}/node_modules/**"
+      ]
     },
     {
-      "name": "Debug Server",
+      "name": "Debug All Tests (Node)",
```
thiagomini, line 35: You can run this script in the Debug window, set a breakpoint in any test file, and it'll stop there correctly

thiagomini (commented):
@oskardudycz  I realized that Node version 21 was causing some issues with the debugger. We were using that mainly to get the Node's test runner glob pattern capabilities, but since I managed to fix that using the node-glob package, we can revert back to v20. Can you try out debugging again? Now it works for me from the debugger terminal and the debugger runner option

.vscode/launch.json
oskardudycz: Works for me now, thank you!

.github/workflows/build_and_test.yml
oskardudycz: Nice finding, thank you for doing the investigation 🙂

oskardudycz (approved):
@thiagomini, thank you a lot for doing all this and doing the detective work!

Added placeholders for int and e2e tests

I added placeholders for int and e2e tests. It seems that `node-glob`, when it's unable to resolve files, is not passing anything to the command. That causes all tests to be run without the pattern. Thus, if you run `npm run build` and you have a dist folder, the script will fail. I'll remove those placeholders when I have the first int and e2e tests.

thiagomini:
@oskardudycz sure, no problem, I'm happy to help! Thanks for the fast feedback and fixes as well! I'll be watching the repo and hopefully contribute further to it :)

oskardudycz:
> I'll be watching the repo and hopefully contribute further to it :)

Feel invited! 🙂

PR #24, opened 2024-02-23
Basic fastify package 

mkubasz:
Goal of this PR is to prepare modular fastify package working with Emmett library.

Should contains default options, middleware and logger.

TBD...


stepaniukm (approved):

packages/emmett-fastify/src/index.ts
```diff
@@ -0,0 +1,102 @@
+import Compress from '@fastify/compress';
+import Etag from '@fastify/etag';
+import Form from '@fastify/formbody';
+import closeWithGrace from 'close-with-grace';
+import Fastify, { type FastifyInstance, type FastifyRequest, type FastifyReply } from 'fastify';
+
+const defaultPlugins = [
```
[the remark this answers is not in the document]
mkubasz: I wouldn't go with flags. Imagine that you need 10 flags to sets. In my opinion will be better to choose some what you want. So rewrite existing setting using filter. But yeah worth to add something like name to filter by some array of names.

packages/emmett-fastify/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+];
+
+const defaultPostMiddlewares = (app: FastifyInstance) => {
+  app.all('*', async (request: FastifyRequest, reply: FastifyReply) => {
+    return reply.status(404).send();
+  });
+};
+
+export interface ApplicationOptions {
+  serverOptions?: { logger: boolean };
```
oskardudycz, line 24: **SUGGESTION:** I think that we should allow people to provide their own server options and just set up the defaults to make configuration smoother. I wouldn't like to wrap the whole initialisation as then each change in fastify will trigger the need to adjust the `ApplicationOptions`.

packages/emmett-fastify/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  registerRoutes?: (app: FastifyInstance) => void;
+  registerPreMiddlewares?: (app: FastifyInstance) => void;
```
oskardudycz, line 26: **SUGGESTION:** Unless Fastify expects to have middleware registered in a certain order (e.g. before adding hooks, then I'd prefer not to have that middleware and extends here. The' getApplication' intends to provide the safe defaults and the most common configuration, but allow users to use native way of configuring tooling. So I'd prefer to just pass the settings that Emmett will be responsible for, so let user do something like:


````ts
const application = getApplication({apis: [shoppingCartApi(eventStore, getUnitPrice, () => now)]});

// do additional registration here
fastify.use(require('something')())
````

File: packages/emmett-fastify/src/index.ts [hunk 4]
```diff
@@ -0,0 +1,101 @@
...
+} from 'fastify';
+
+const defaultPlugins = [
+  { plugin: Etag, options: {} },
+  { plugin: Compress, options: { global: false } },
+  { plugin: Form, options: {} },
+];
+
+const defaultPostMiddlewares = (app: FastifyInstance) => {
+  app.all('*', async (request: FastifyRequest, reply: FastifyReply) => {
```
oskardudycz, line 18: **CONCERN:** Shouldn't we use `setNotFoundHandler` here? https://fastify.dev/docs/latest/Reference/Server/#setnotfoundhandler

Also shouldn't we have here also EmmetError mapping? I think that Fastify maps automatically `statusCode` field from error, and EmmettError has `errorCode` instead, so I think that it won't be automatically mapped.

packages/emmett-fastify/src/index.ts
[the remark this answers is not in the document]
oskardudycz: I'm fine for now with weakly typed, but the feedback by @stepaniukm is valid. We should consider making it strongly typed. I'll also need to consider the general approach to plugins in Emmett. 

For the context, I used flags intentionally as scarification. They're so ugly that if someone disables the default flags, then the bell should already ring, and they should ensure that's something they should do carefully. Still, of what I saw plugins should work for as it's aligned with Fastify conventions.

packages/emmett-fastify/src/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  });
+};
+
+export interface ApplicationOptions {
+  serverOptions?: { logger: boolean };
+  registerRoutes?: (app: FastifyInstance) => void;
+  registerPreMiddlewares?: (app: FastifyInstance) => void;
+  registerPostMiddlewares?: (app: FastifyInstance) => void;
+  extendApp?: (app: FastifyInstance) => FastifyInstance;
+  activeDefaultPlugins?: Array<{
```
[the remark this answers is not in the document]
oskardudycz: Good call; still, I'm fine with handling that in the follow-up PR.

packages/emmett-fastify/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const closeListeners = closeWithGrace({ delay: 500 }, async (opts) => {
+    if (opts.err) {
+      app.log.error(opts.err);
+    }
+
+    await app.close();
+  });
+
+  app.addHook('onClose', (instance, done) => {
```
oskardudycz, line 75: **KUDOS:** Nice, I need to add that also to Express :+1:

oskardudycz (commented):
@mkubasz I added some comments/concerns/suggestions. Could you have a look at them? (and also do the rebase :wink: ).

Do you know if Fastify supports Problem Details out of the box?

Thanks for coming up with this PR; I think that once we agree on those comments and suggestions, we could pull it in.

packages/emmett-fastify/src/index.ts
mkubasz: I will remove this to allow user use handling error how they want.

oskardudycz (approved):
I'm fine with it as it is. We can tackle remaining items like problem details, integration and e2e test separately.

Btw. @mkubasz for the future, could you do rebase instead of merge? Then I won't need to do squash before merging :)

oskardudycz:
@mkubasz I added commit with fixes to Fastify API tests and typing around plugins. I also fixed linter errors. See details in: https://github.com/event-driven-io/emmett/pull/24/commits/43a2aae03ad79b851e3b7a5ae275e48a4a383778

PR #27, opened 2024-03-01
Added implementation of the ESDB Event Store 🐲

oskardudycz:
Tests will come after the follow-up PR with the TestContainers package.


oskardudycz:
@watfordsuzy, you may be interested in the EventStoreDB implementation. It has full OC handling implemented 🙂

PR #29, opened 2024-03-02
feat(docs,emmet-expressjs): add e2e api specification

### LINKED ISSUE #26: Add ability to setup integration tests through HTTP requests
Currently, integration test data can be set up only by providing the initial set of events. That should be enough for most cases, but some people prefer to do it fully through the Web API, treating the application as a black box.

The entry point of that is located here: https://github.com/event-driven-io/emmett/blob/aa4413fceb90a6bcdba88823456ad8fba9842186/packages/emmett-expressjs/src/testing/apiSpecification.ts#L87.

It should be done in a similar way as with asserts but allowing to pass multiple setups. Setup should use the same structure as when passing the `supertest` setup https://github.com/event-driven-io/emmett/blob/aa4413fceb90a6bcdba88823456ad8fba9842186/packages/emmett-expressjs/src/testing/apiSpecification.ts#L89.

thiagomini:
Closes #26

Hey @oskardudycz , I haven't added the ES db yet because I'm unsure how to replace the `inMemoryStore()`. However, I believe this draft may suffice as an initial discussion for that issue. It took me some time to figure out how everything worked and glued together, but I feel I'm understanding the architecture better now 😅

One thing I'd like to ask / discuss is the [no-unsafe-assignement](https://typescript-eslint.io/rules/no-unsafe-assignment/) rule that we're using here and it's complaining all over the place - is this intended? Honestly, in many cases it doesn't make sense, look at this example:

image

The `getInMemoryEventStore()` function has a defined return value, yet this rule is complaining 🤔 

Anyway, let me know what you think about this initial implementation!

### DIFF
--- a/docs/snippets/gettingStarted/webApi/apiBDD.e2e.spec.ts
+++ b/docs/snippets/gettingStarted/webApi/apiBDD.e2e.spec.ts
@@ -1,0 +1,77 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+import {
+  getInMemoryEventStore,
+  type EventStore,
+} from '@event-driven-io/emmett';
+import {
+  ApiE2ESpecification,
+  expectResponse,
+  getApplication,
+} from '@event-driven-io/emmett-expressjs';
+import { randomUUID } from 'node:crypto';
+import { beforeEach, describe, it } from 'node:test';
+import type { PricedProductItem } from '../events';
+import { ShoppingCartStatus } from './shoppingCart';
+import { shoppingCartApi } from './simpleApi';
+
+const getUnitPrice = () => {
+  return Promise.resolve(100);
+};
+
+describe('ShoppingCart E2E', () => {
+  let clientId: string;
+  let shoppingCartId: string;
+
+  beforeEach(() => {
+    clientId = randomUUID();
+    shoppingCartId = `shopping_cart:${clientId}:current`;
+  });
+
+  describe('When empty', () => {
+    it('should add product item', () => {
+      return given((request) =>
+        request
+          .post(`/clients/${clientId}/shopping-carts/current/product-items`)
+          .send(productItem),
+      )
+        .when((request) =>
+          request.get(`/clients/${clientId}/shopping-carts/current`).send(),
+        )
+        .then([
+          expectResponse(200, {
+            body: {
+              clientId,
+              id: shoppingCartId,
+              productItems: [
+                {
+                  quantity: productItem.quantity,
+                  productId: productItem.productId,
+                },
+              ],
+              status: ShoppingCartStatus.Opened,
+            },
+          }),
+        ]);
+    });
+  });
+
+  const now = new Date();
+
+  const given = ApiE2ESpecification.for(
+    (): EventStore => getInMemoryEventStore(),
+    (eventStore: EventStore) =>
+      getApplication({
+        apis: [shoppingCartApi(eventStore, getUnitPrice, () => now)],
+      }),
+  );
+
+  const getRandomProduct = (): PricedProductItem => {
+    return {
+      productId: randomUUID(),
+      unitPrice: 100,
+      quantity: Math.random() * 10,
+    };
+  };
+
+  const productItem = getRandomProduct();
+});
--- a/docs/snippets/gettingStarted/webApi/shoppingCart.ts
+++ b/docs/snippets/gettingStarted/webApi/shoppingCart.ts
@@ -52,6 +52,7 @@ export enum ShoppingCartStatus {
   Pending = 'Pending',
   Confirmed = 'Confirmed',
   Canceled = 'Canceled',
+  Opened = 'Opened',
 }
 
 export type Empty = { status: ShoppingCartStatus.Empty };
--- a/docs/snippets/gettingStarted/webApi/simpleApi.ts
+++ b/docs/snippets/gettingStarted/webApi/simpleApi.ts
@@ -6,6 +6,7 @@ import {
 } from '@event-driven-io/emmett';
 import {
   NoContent,
+  OK,
   on,
   type WebApiSetup,
 } from '@event-driven-io/emmett-expressjs';
@@ -22,7 +23,8 @@ import type {
   ConfirmShoppingCart,
   RemoveProductItemFromShoppingCart,
 } from '../commands';
-import { evolve, getInitialState } from '../shoppingCart';
+import type { ProductItem, ShoppingCartEvent } from '../events';
+import { evolve, getInitialState, type ShoppingCart } from '../shoppingCart';
 
 export const handle = CommandHandler(evolve, getInitialState);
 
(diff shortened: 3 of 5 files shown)


docs/snippets/gettingStarted/webApi/apiBDD.e2e.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+const getUnitPrice = () => {
+  return Promise.resolve(100);
+};
+
+describe('ShoppingCart E2E', () => {
+  let clientId: string;
+  let shoppingCartId: string;
+
+  beforeEach(() => {
+    clientId = randomUUID();
```
thiagomini, line 26: Notice that we don't need the `uuid()` library to generate uuid v4 - it's already supported by node's native "crypto" module

docs/snippets/gettingStarted/webApi/apiBDD.e2e.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  describe('When empty', () => {
+    it('should add product item', () => {
+      return given((request) =>
+        request
+          .post(`/clients/${clientId}/shopping-carts/current/product-items`)
+          .send(productItem),
+      )
+        .when((request) =>
+          request.get(`/clients/${clientId}/shopping-carts/current`).send(),
```
thiagomini, line 38: I've created this temporal workaround `get` method - I thought that it would be good to do this and in the future we just need to change the underlying implementation detail to use projections

docs/snippets/gettingStarted/webApi/shoppingCart.ts
```diff
@@ -52,6 +52,7 @@ export enum ShoppingCartStatus {
   Pending = 'Pending',
   Confirmed = 'Confirmed',
   Canceled = 'Canceled',
+  Opened = 'Opened',
```
thiagomini, line 55: This was necessary so I could check for it in the e2e test

packages/emmett-expressjs/src/e2e/decider/applicationLogicWithOC.int.spec.ts
```diff
@@ -44,7 +44,6 @@ describe('Application logic with optimistic concurrency', () => {
 
     if (!current.id) {
       assert.fail();
-      return;
```
thiagomini, line 47: when we call `assert.fail()`, node automatically throws an assertion error, so this line was unreachable

packages/emmett-expressjs/src/testing/apiSpecification.ts
```diff
@@ -59,9 +61,9 @@ export const expectNewEvents = <EventType extends Event = Event>(
 };
 
 export const expectResponse =
-  (
+  <Body = unknown>(
```
thiagomini, line 64: A small addition to the `expectResponse` matcher so we can optionally type the body

docs/snippets/gettingStarted/webApi/simpleApi.ts
```diff
@@ -139,6 +141,40 @@ export const shoppingCartApi =
         return NoContent();
       }),
     );
+
```
oskardudycz, line 144: **FIX:** @thiagomini, for now I'd prefer to keep the sample outside of the `complete-api` region as the example in the getting started is (for now) focused on business logic and writing. I'm planning to introduce the read model side as the next step. Would you mind moving this endpoint outside the region? 

The method as a test method is okay as it is. You could consider providing the customised evolve method that would already build the desired read model without the need to do the additional mapping. But that's a nice to have if you wanted to play more with event sourcing. This would be a dedicated `ShoppingCartDetails` type with is own `evolve` function.

docs/snippets/gettingStarted/webApi/simpleApi.ts
```diff
@@ -139,6 +141,40 @@ export const shoppingCartApi =
...
+          productId,
+          quantity,
+        }));
+
+        return OK({
+          body: {
+            clientId: assertNotEmptyString(request.params.clientId),
+            id: shoppingCartId,
+            productItems,
+            status: result?.state.status,
```
oskardudycz, line 173: **SUGGESTION:** I think that instead of using `?` here, you could return `NotFound` status if result is `null`

docs/snippets/gettingStarted/webApi/simpleApi.ts
```diff
@@ -139,6 +141,40 @@ export const shoppingCartApi =
         return NoContent();
       }),
     );
+
+    // Get Shopping Cart
+    router.get(
```
oskardudycz, line 146: **FIX:** @thiagomini, for now I'd prefer to keep the sample outside of the `complete-api` region as the example in the getting started is (for now) focused on business logic and writing. I'm planning to introduce the read model side as the next step. Would you mind moving this endpoint outside the region?

oskardudycz, line 146: **HINT:** The method as a test method is okay as it is. You could consider providing the customised evolve method that would already build the desired read model without the need to do the additional mapping. But that's a nice to have if you wanted to play more with event sourcing. This would be a dedicated `ShoppingCartDetails` type with is own `evolve` function.

packages/emmett-expressjs/src/e2e/decider/applicationLogicWithOC.int.spec.ts
oskardudycz: ah, didn't know that. Thanks for the hint!

packages/emmett-expressjs/src/testing/apiSpecification.ts
oskardudycz: Makes perfect sense to me, thanks for doing that! 🫡

docs/snippets/gettingStarted/webApi/apiBDD.e2e.spec.ts
oskardudycz: Ah, cool. For those needs, this will be sufficient. Happy to remove the dependency on `uuid` package!

docs/snippets/gettingStarted/webApi/apiBDD.e2e.spec.ts
oskardudycz: Yes, sounds good to me, I like the idea. Thanks to that, we'll have the full coverage.

oskardudycz (approved):
@thiagomini it looks good to me even as it is. Check my comments, if you'd like to apply some of them. I added the EventStoreDB config for e2e tests, see details in my commit :slightly_smiling_face:

packages/emmett-expressjs/src/testing/apiSpecification.ts
```diff
@@ -151,6 +161,50 @@ export const ApiSpecification = {
   },
 };
 
+export const ApiE2ESpecification = {
```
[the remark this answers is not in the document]
thiagomini: Sure, I can do that!

packages/emmett-expressjs/src/testing/apiSpecification.ts
```diff
@@ -123,16 +113,16 @@ export const ApiSpecification = {
...
+                  const succeeded = verify(response);
 
-                  if (succeded === false) assert.fail();
+                  if (succeeded === false) assert.fail();
                 } else if (Array.isArray(verify)) {
                   const [first, ...rest] = verify;
 
                   if (typeof first === 'function') {
-                    const succeded = first(response);
+                    const succeeded = first(response);
```
oskardudycz, line 123: @thiagomini, thanks for fixing my typos! 😅

thiagomini: Sure, no problem haha I only got them because I have the "Code Spell Cheker" extension in my VSCode

PR #30, opened 2024-03-02
Fix martendb.io link in README.md

futpib:
Now it points to https://github.com/event-driven-io/emmett/blob/main/martendb.io (404)


oskardudycz (approved):
Thank you!

PR #36, opened 2024-03-07
Removed project references from main tsconfig

### LINKED ISSUE #18: Fix linter for global config
It's behaving really weird, working locally but not in gh action, see https://github.com/event-driven-io/emmett/actions/runs/7988114136/job/21812011212. Sounds like some config is wrong. Tried setting up projects references explicitly to tsconfigs but didn't work out.

oskardudycz:
Needed to configure a separate config for eslint, as it doesn't deal well with the TypeScript project references. Also, it wasn't checking correctly the eslint settings, as eslint requires wrapping global pattern in `''` to be passed as a parameter instead of taken as the shell pattern. :facepalm: 

Updated also pipeline to run ts build before linting to have correctly resolved dependencies between packages and non-package code.

Fixes #18.


.eslintrc.json
```diff
@@ -11,12 +11,12 @@
     "plugin:@typescript-eslint/recommended-requiring-type-checking",
     "plugin:prettier/recommended"
   ],
-  "ignorePatterns": ["dist/", "lib/"],
+  "ignorePatterns": ["**/dist/", "**/lib/"],
```
oskardudycz, line 14: To ignore nested folders inside packages.

packages/emmett-esdb/tsconfig.eslint.json
```diff
@@ -0,0 +1,3 @@
+{
```
oskardudycz, line 1: Those dummy files are needed for vscode eslint plugin, as it runs eslint config locally for packages. And as in the file, we have a reference for `tsconfig.eslint.json`. Then if it won't find such a file in the local package folder, then it will cry out loud and doesn't highlight eslint errors.

PR #51, opened 2024-03-18
Feat/create testing abstraction

### LINKED ISSUE #34: Test: Devise a testing strategy that we can use to reduce duplication among emmet packages
# Description

As we noted in this [PR](https://github.com/event-driven-io/emmett/pull/33), some test cases are duplicated among our packages. Therefore, when we want to fix them or change something, we have to repeat the same code in multiple files, which is unnecessary and prone to human error. We need to create some sort of matrix testing **or** a testing utility that can be reused across packages to execute the same logical test.

A few things to discuss:

- Each of these "duplicated" tests are located in different packages. For instance, [inMemoryTest](https://github.com/event-driven-io/emmett/blob/main/packages/emmett/src/eventStore/inMemoryEventStore.unit.spec.ts) in `emmet` package and [eventStoreTest](https://github.com/event-driven-io/emmett/blob/main/packages/emmett-esdb/src/eventStore/eventstoreDBEventStore.e2e.spec.ts) in the `emmet-esdb` one. Should we keep them separated?
- What is common between these tests, and what's different? What can we abstract away to facilitate creating tests for future integrations?

## Acceptance Criteria

1. Definition and agreement of a testing strategy that can reduce the boilerplate / duplicate code between tests
2. Implementation of the aforementioned strategy in existing tests

## Additional Information

To inspire this, we can use [ESLint's Testing Helper](https://eslint.org/docs/latest/integrate/nodejs-api#ruletester). They've created a tool to facilitate creating tests that are usually very similar.

thiagomini:
This should close #34 once ready for merge.

The idea here is to create some utility function that we can use to create similar tests between different Event Store implementations. Currently, we've done two major changes:

1. Extracted the common business rule code to the `shoppingCart.domain` file
2. Created an example utility function under `emmet/src/testing/features.ts` file.

Alright, this should already be working for existing tests, @oskardudycz . Now, a few things to discuss:

1. Where should we put the common business logic used by all tests? Currently it's in `src/packages/emmett/src/testing/shoppingCart.domain.ts`
2. Should we consider creating a separate package like `emmet-testing` where the common test utility resides? For instance, we could put both this "domain" logic and the function utilities to test new features (which currently are in `src/packages/emmett/src/testing/features.ts`
3. Should we make these test utilities more abstract? As you can see, I've created the `testAggregateStream()` function which is specifically for testing aggregate stream features. However, we certainly will need to test more features in the future, and we should consider how much duplication we want.

### DIFF
--- a/src/packages/emmett/src/eventStore/inMemoryEventStore.unit.spec.ts
+++ b/src/packages/emmett/src/eventStore/inMemoryEventStore.unit.spec.ts
@@ -7,5 +7,5 @@ import { testAggregateStream } from '../testing';
 
 describe('InMemoryEventStore', () => {
   const eventStore = getInMemoryEventStore();
-  testAggregateStream(eventStore);
+  testAggregateStream(() => Promise.resolve(eventStore));
 });
--- a/src/packages/emmett/src/eventStore/inMemoryEventStore.unit.spec.ts
+++ b/src/packages/emmett/src/eventStore/inMemoryEventStore.unit.spec.ts
@@ -1,80 +1,11 @@
 /* eslint-disable @typescript-eslint/no-floating-promises */
-import assert from 'node:assert';
-import { randomUUID } from 'node:crypto';
-import { describe, it } from 'node:test';
+import { describe } from 'node:test';
 import { getInMemoryEventStore } from '../eventStore';
-import {
-  evolve,
-  getInitialState,
-  type PricedProductItem,
-  type ShoppingCartEvent,
-} from '../testing/shoppingCart.domain';
+import { testAggregateStream } from '../testing';
 
 // Events & Entity
 
 describe('InMemoryEventStore', () => {
   const eventStore = getInMemoryEventStore();
-
-  describe('aggregateStream', () => {
-    it('When called with `to` allows time traveling', async () => {
-      // Given
-      const productItem: PricedProductItem = {
-        productId: '123',
-        quantity: 10,
-        price: 3,
-      };
-      const discount = 10;
-      const shoppingCartId = randomUUID();
-
-      await eventStore.appendToStream<ShoppingCartEvent>(shoppingCartId, [
-        { type: 'ProductItemAdded', data: { productItem } },
-      ]);
-      await eventStore.appendToStream<ShoppingCartEvent>(shoppingCartId, [
-        { type: 'ProductItemAdded', data: { productItem } },
-      ]);
-      await eventStore.appendToStream<ShoppingCartEvent>(shoppingCartId, [
-        { type: 'DiscountApplied', data: { percent: discount } },
-      ]);
-
-      // when
-      const resultAt1 = await eventStore.aggregateStream(shoppingCartId, {
-        evolve,
-        getInitialState,
-        read: { to: 1n },
-      });
-      const resultAt2 = await eventStore.aggregateStream(shoppingCartId, {
-        evolve,
-        getInitialState,
-        read: { to: 2n },
-      });
-      const resultAt3 = await eventStore.aggregateStream(shoppingCartId, {
-        evolve,
-        getInitialState,
-        read: { to: 3n },
-      });
-
-      // then
-      assert.ok(resultAt1);
-      assert.ok(resultAt2);
-      assert.ok(resultAt3);
-
-      assert.equal(resultAt1.currentStreamVersion, 1);
-      assert.deepEqual(resultAt1.state, {
-        productItems: [productItem],
-        totalAmount: 30,
-      });
-
-      assert.equal(resultAt2.currentStreamVersion, 2);
-      assert.deepEqual(resultAt2.state, {
-        productItems: [productItem, productItem],
-        totalAmount: 60,
-      });
-
-      assert.equal(resultAt3.currentStreamVersion, 3);
-      assert.deepEqual(resultAt3.state, {
-        productItems: [productItem, productItem],
-        totalAmount: 54,
-      });
-    });
-  });
+  testAggregateStream(eventStore);
 });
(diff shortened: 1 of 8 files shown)


src/packages/emmett/src/testing/features.ts
```diff
@@ -0,0 +1,90 @@
...
+type TestOptions = {
+  getInitialIndex: () => bigint;
+};
+
+export type EventStoreFactory = () => Promise<{
+  eventStore: EventStore<bigint>;
+  tearDown?: () => Promise<void>;
+}>;
+
+export async function testAggregateStream(
```
thiagomini, line 22: I still have to work on this. It only works for readily available Event Stores that don't need a setup or teardown phase.

src/packages/emmett/src/eventStore/inMemoryEventStore.unit.spec.ts
```diff
@@ -1,115 +1,11 @@
...
-      });
-
-      assert.equal(resultAt3.currentStreamVersion, 3);
-      assert.deepEqual(resultAt3.state, {
-        productItems: [productItem, productItem],
-        totalAmount: 54,
-      });
-    });
-  });
+  testAggregateStream(eventStore);
```
thiagomini, line 10: The idea would be to have these features defined in the `testing` folder, and each implementation would call it with the appropriate settings. Therefore, we only have to implement the test logic once

src/packages/emmett/src/testing/shoppingCart.feature
```diff
@@ -0,0 +1,17 @@
+Feature: ShoppingCart
```
oskardudycz, line 1: **QUESTION:** Out of curiosity, is this the way how you typically document testing scenarios, or is it also some longer shot behind it (e.g. DSL for testing)? Disclaimer, I really like that you documented it!

oskardudycz:
@thiagomini, all of that looks great to me, that's more or less what I was considering. Here are my comments and answer to your point:

> 1. Where should we put the common business logic used by all tests? Currently it's in src/packages/emmett/src/testing/shoppingCart.domain.ts

> 2. Should we consider creating a separate package like emmet-testing where the common test utility resides? For instance, we could put both this "domain" logic and the function utilities to test new features (which currently are in src/packages/emmett/src/testing/features.ts

We could add a private package, which will be a development-time-only reference for others. Then we won't need to repeat those common scenarios. We could also put there code for those internal test engines that we don't necessarily want to expose. 

For the code that we'd like to expose to the users, I think that it is better to remain in the main package and then decrease the friction of adding a new package. I'm already excluding `.spec` files from production build https://github.com/event-driven-io/emmett/blob/9e83e9bf1d7878a2d00c2ff238907de481b0292b/src/packages/emmett/tsup.config.ts#L17. 

Of course, we need to be careful in that, as having some clear implementation is beneficial to see the stub of what we're testing. 

Having this testing package could be useful if, at some point, someone wanted to implement their own store that we wouldn't like to support officially; then, we could expose this private package and make it easier to run compatibility tests. But I prefer to expose less and then make it public as there's a real need. It's easier to handle breaking changes by doing that.

Thoughts?

> 3. Should we make these test utilities more abstract? As you can see, I've created the testAggregateStream() function which is specifically for testing aggregate stream features. However, we certainly will need to test more features in the future, and we should consider how much duplication we want.

For now, indeed, having the granular tests seems a bit odd, but I think that as time goes on, different implementations of the event store may diverge in terms of behaviour (e.g. might supporting a subset of features or implementing some guarantees differently). Then I think that's worth keeping them as you did. We can even keep each scenario/group of tests (e.g. stream aggregation) in separate files. That will make maintenance easier. It'll also allow adding features gradually, so implementing event store step by step and adding tests unless it matches all scenarios. We can, of course, provide some wrapper for now that will run all tests (so no one forgets to add some tests).

Thoughts?

src/packages/emmett/src/eventStore/inMemoryEventStore.unit.spec.ts
oskardudycz: Makes sense to me :+1:

src/packages/emmett/src/testing/shoppingCart.feature
thiagomini: Hey @oskardudycz! Well, this is typically how I do it when working on a product - I first write down the acceptance criteria of a task like this. However, I did it here to clarify the intent behind these "Shareable" tests. But I'm concerned about ensuring it's up-to-date with the implementation - maybe we should convert it to an internal executable DSL - thoughts?

thiagomini:
Hey @oskardudycz , thanks for the review! Regarding the creation of a new test package, would you want me to do it already? This is still on its early stages, and maybe we can postpone it until we have more info on how testing will look like in the future 🤔 

Regarding the granularity of the `testAggregateStream()` function, it will have to change in the future, but I agree we can keep it as it is for now. I fear that if we try to abstract too much too early, we'll end up with an overly complex mess. So, my final question would be, do you think this PR needs any changes now, or can we merge it?

src/packages/emmett/src/testing/shoppingCart.feature
oskardudycz: I was thinking initially about generating human-readable markdown or DSL from tests at some point, but we can do it either way. You mentioned that you did some DSL-based tests in Node. We can discuss that as a separate issue. I’m fine with having them as they are 🙂

Thank you for clarifications 👍

oskardudycz:
@thiagomini thanks for your thoughts. I’m fine taking it as it is and then evolving. I’m all for gradual improvement instead of long-living branches 🙂

oskardudycz (approved):

Issue #59, opened 2024-04-19
Missing Dependency `test` when attempting to use `@event-driven-io/emmett`


alex-laycalvert:
When attempting to use Emmett (originally found when running tests with Jest but occurs when importing from Emmett at all) I keep getting the below error:

Node: v21.7.2 (happened for v20 and I'm assuming all other versions)
Emmett: v0.7.0

```
$ node ./index.js

node:internal/modules/cjs/loader:1145
  throw err;
  ^

Error: Cannot find module 'test'
Require stack:
- /home/alex/git/emmett-jest-example/node_modules/@event-driven-io/emmett/dist/chunk-6WITJBGG.js
- /home/alex/git/emmett-jest-example/node_modules/@event-driven-io/emmett/dist/index.js
- /home/alex/git/emmett-jest-example/index.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1142:15)
    at Module._load (node:internal/modules/cjs/loader:983:27)
    at Module.require (node:internal/modules/cjs/loader:1230:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/home/alex/git/emmett-jest-example/node_modules/@event-driven-io/emmett/dist/chunk-
6WITJBGG.js:1:348)
    at Module._compile (node:internal/modules/cjs/loader:1368:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1426:10)
    at Module.load (node:internal/modules/cjs/loader:1205:32)
    at Module._load (node:internal/modules/cjs/loader:1021:12)
    at Module.require (node:internal/modules/cjs/loader:1230:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/alex/git/emmett-jest-example/node_modules/@event-driven-io/emmett/dist/chunk-6WITJBGG.js',
    '/home/alex/git/emmett-jest-example/node_modules/@event-driven-io/emmett/dist/index.js',
    '/home/alex/git/emmett-jest-example/index.js'
  ]
}

Node.js v21.7.2
```

Looked into the issue and looks like a missing dependency of `test`. Will probably have a PR up soon but wanted to document.

alex-laycalvert:
Confirmed installing `test` fixed the issue

alex-laycalvert:
Link to minimum example: https://github.com/alex-laycalvert/emmett-jest-error

PR #60, opened 2024-04-19
Added `node:test` as a dependency to `emmett` to prevent `MODULE_NOT_FOUND` errors

alex-laycalvert:
Saved `test` as a dependency in the `emmett` package. Not sure if the files that use `test` were meant to be in the final build but this solved my issues locally after finding #59


alex-laycalvert:
This might need to be a peer dep, I've never done any work on node libraries so not sure which is better.

oskardudycz:
@alex-laycalvert, what you did is a valid solution, but I decided to tackle that issue a bit differently. More details in https://github.com/event-driven-io/emmett/pull/61. 

Thank you a lot for providing good reproduce steps, a potential fix and patience. That's much appreciated. I'll go ahead and publish the 0.7.1 version with a fix now.

PR #76, opened 2024-06-12
Added streaming capabilities

oskardudycz:
This PR introduces foundations for streaming capabilities based on web streams. It adds shim for using native web streams or Node.js and stream transformations.

This PR has not yet introduced the streaming API; it'll come in the follow-up PR. I intended to deliver it here, but this PR was already:

neverending story

So let's merge it and finally introduce the final usage in a separate PR.


oskardudycz:
the end is near

PR #85, opened 2024-07-08
Updated missing link in getting-started.md

d3spis3d:
### Context

Thanks for sharing this great project. I was working through the Getting Started page and noticed a missing link.

### What's Changed

I searched for what I believe is the correct article being referenced and updated the link in `getting-started.md`. Let me know if I've referenced the wrong article.


oskardudycz (approved):
Thank you!

PR #90, opened 2024-07-12
Added a first experimental implementation of PostgreSQL inline projections

oskardudycz:


src/packages/emmett-postgresql/src/eventStore/projections/index.ts
```diff
@@ -0,0 +1,41 @@
...
+export type PostgresProjectionHandler<EventType extends Event = Event> = (
+  connectionString: string,
+  client: pg.PoolClient,
+  events: ReadEvent<EventType>[],
+) => Promise<void> | void;
+
+export type ProjectionDefintion<EventType extends Event = Event> = {
+  type: 'inline';
+  name?: string;
+  canHandle: EventTypeOf<EventType>[];
```
dpraimeyuu, line 18: @oskardudycz just wondering, does it make sense to have a projection that can handle no events?

Shouldn't it be `NonEmptyArrayOf<EventTypeOf<EventType>>` :thinking: ?

src/packages/emmett-postgresql/src/eventStore/projections/pongo.ts
```diff
@@ -0,0 +1,92 @@
...
+  eventHandler: PongoProjectionHandler<EventType>;
+  eventTypes: EventTypeOf<EventType>[];
+};
+
+export type PongoProjectionHandler<EventType extends Event = Event> = (
+  documentId: (event: ReadEvent<EventType>) => string,
+  connectionString: string,
+  client: pg.PoolClient,
+  events: ReadEvent<EventType>[],
+) => Promise<void> | void;
```
dpraimeyuu, line 25: @oskardudycz Not caring about allocation, would it make sense to wrap all arguments in an object to make them position-independent?

PR #96, opened 2024-08-05
Exposed custom projection builders and ProjectionDefinition types

oskardudycz:
It appeared that ProjectionDefinition types were not exposed and that caused issues with setting up custom projections.

Added builders to make that easier.

Reshaped also ProjectionDefnition helper method moving connection string and client to second context parameter. Thanks to that users that don't need it will have a simpler handle method definition.


oskardudycz:
@ jameswoodley FYI

PR #101, opened 2024-08-07
Refactored PostgreSQL event store to benefit from the latest Pongo and Dumbo improvements

oskardudycz:
Now, it's possible to inject ambient connections and use non-pooled event stores in the same way as in Pongo.

To run a non-pooled setup (e.g. for Cloudflare Hyperdrive or Supabase Supavisor) you need to do the following:

```ts
const eventStore = getPostgreSQLEventStore(connectionString, {
  projections: [shoppingCartShortInfoProjection, customProjection],
  connectionOptions: { pooled: false },
});
```

It'll also be easier to deliver SQL projections, but that'll come in the follow-up PR.

### DIFF
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@event-driven-io/core",
-  "version": "0.14.1",
+  "version": "0.15.0",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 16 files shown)


oskardudycz:
@jameswoodley FYI

PR #102, opened 2024-08-07
Added raw sql projections to support raw SQL handling

### LINKED ISSUE #97: Add projection builder that takes handler with event and return sql statements

oskardudycz:
Generalised also projection definition.

Made `postgreSQLProjection` and `inlineProjection` obsolete, use `postgreSQLProjection` and `postgreSQLInlineProjection` instead.

Now you can use the new projections raw projection as:

```ts
import { postgreSQLRawSQLProjection } from '@event-driven-io/emmett-postgresql';

const rawSQLProjection = postgreSQLRawSQLProjection<ProductItemAdded>({
  name: 'customProjection',
  canHandle: ['ProductItemAdded'],
  handle: (event, context) => sql('INSERT INTO product_items VALUES(%s, %L, %L)', event.productId, event.quantity))
});
```

There's also `postgreSQLRawBatchSQLProjection` that handles a batch of events accordingly. Both can be sync or async. You can access and query the database through the `context` handler param.

Fixes #97

### DIFF
--- a/src/packages/emmett-postgresql/src/eventStore/postgreSQLEventStore.ts
+++ b/src/packages/emmett-postgresql/src/eventStore/postgreSQLEventStore.ts
@@ -24,9 +24,9 @@ import {
 } from '@event-driven-io/emmett';
 import pg from 'pg';
 import {
-  defaultProjectionOptions,
+  defaultPostgreSQLProjectionOptions,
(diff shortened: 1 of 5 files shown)


oskardudycz:
@jameswoodley FYI

PR #104, opened 2024-08-09
Refactored Postgres projections registrations

oskardudycz:
- Separated projection definition from registration. Thanks to that, projections will be able to switch their type in the future or be used in both (e.g., for rebuilds).
- Refactored projections to take params object instead of a list of parameters. That will cut the number of permutations and make it easier to define them.
- Added reusable type CanHandle to unified Event Handlers typing restrictions.
- Renamed `pongoSingleProjection to `pongoSingleStreamProjection` to keep naming consistency.

Now, registration will look like this (notice `projections.inline` helper):

```ts
const eventStore = getPostgreSQLEventStore(connectionString, {
  projections: projections.inline(
    [shoppingCartShortInfoProjection, customProjection]
  ),
});
````

And Pongo projection definition as:

````ts
const shoppingCartShortInfoProjection = pongoSingleStreamProjection({
  collectionName: 'shoppingCartsShortInfo',
  evolve,
  canHandle: ['ProductItemAdded', 'DiscountApplied'],
});
````

### DIFF
--- a/src/packages/emmett-postgresql/src/eventStore/postgreSQLEventStore.e2e.spec.ts
+++ b/src/packages/emmett-postgresql/src/eventStore/postgreSQLEventStore.e2e.spec.ts
@@ -27,7 +27,7 @@ import {
   type PostgresEventStore,
 } from './postgreSQLEventStore';
 import { postgreSQLProjection } from './projections';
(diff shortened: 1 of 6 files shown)


oskardudycz:
@jameswoodley FYI

PR #125, opened 2024-10-25
Removed reference to commander in a regular (non-CLI) package usage

### LINKED ISSUE #74: Add tests for Emmett ensuring browser compatibility

### LINKED ISSUE #124: Browser Only
Not sure if this is still in scope for Emmet, but running in browser only for prototyping is not supported with Emmet 0.20.x anymore, as it seems the eventstore relies on the commander-library which is using EventEmitter under the hood, which requires Node.

oskardudycz:
Made plugin code unrelated to `commander` package. Thanks to that even if Emmett has `commander` peer dependency, it'll be only loaded for Command Line usage (when `@event-driven-io/emmett/cli` is imported.

Added test to detect future regression. See:
- [Failed build without the fix](https://github.com/event-driven-io/emmett/actions/runs/11517613085/job/32062648285)
- [Successful build with the fix](https://github.com/event-driven-io/emmett/actions/runs/11517701193/job/32062921413) 

Fixes #124 
Fixes #74

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -83,13 +83,13 @@ jobs:
             echo "No errors found, proceeding..."
           fi
 
-      # - name: Use Emmett shims in the browser compatibility test project
-      #   working-directory: ./src/e2e/browser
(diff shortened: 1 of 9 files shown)


dilgerma:
thanks!

PR #126, opened 2024-10-26
(WIP) MongoDB (w/ `mongoose`) Event Store Implentation

alex-laycalvert:
Work in progress PR for an `EventStore` implementation for `mongodb`.

I am utilizing [mongoose](https://mongoosejs.com/docs/) to assist in the model definition but this is only as a wrapper for the actual mongodb driver. If we don't want to use `mongoose` I can look at removing it's usage.

Another thing that might need to change is using `number` for the default stream version instead of a `bigint`. I only did this to make dealing with the version checking easier. In this impl., the stream version is determined by the number of events which is an array field on the document representing each event stream and that `.length` field is _always_ a number type.

@oskardudycz looking forward on working on this one with you and getting this in once it's fully ready

### DIFF
--- a/src/packages/emmett-mongodb/package.json
+++ b/src/packages/emmett-mongodb/package.json
@@ -1,0 +1,58 @@
+{
+  "name": "@event-driven-io/emmett-mongodb",
+  "type": "module",
+  "version": "0.20.2",
+  "description": "Emmett - MongoDB - Event Sourcing development made simple",
+  "scripts": {
+    "build": "tsup",
+    "build:ts": "tsc",
+    "build:ts:watch": "tsc  --watch",
+    "test": "run-s test:unit test:int test:e2e",
+    "test:unit": "glob -c \"node --import tsx --test\" **/*.unit.spec.ts",
+    "test:int": "glob -c \"node --import tsx --test\" **/*.int.spec.ts",
+    "test:e2e": "glob -c \"node --import tsx --test\" **/*.e2e.spec.ts",
+    "test:watch": "node --import tsx --test --watch",
+    "test:unit:watch": "glob -c \"node --import tsx --test --watch\" **/*.unit.spec.ts",
+    "test:int:watch": "glob -c \"node --import tsx --test --watch\" **/*.int.spec.ts",
+    "test:e2e:watch": "glob -c \"node --import tsx --test --watch\" **/*.e2e.spec.ts"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/event-driven-io/emmett.git"
+  },
+  "keywords": [
+    "Event Sourcing"
+  ],
+  "author": "Oskar Dudycz",
+  "bugs": {
+    "url": "https://github.com/event-driven-io/emmett/issues"
+  },
+  "homepage": "https://event-driven-io.github.io/emmett/",
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
+    }
+  },
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "dependencies": {
+    "mongoose": "^8.7.3"
+  },
+  "devDependencies": {
+    "@event-driven-io/emmett-testcontainers": "0.20.2"
+  },
+  "peerDependencies": {
+    "@event-driven-io/emmett": "0.20.2"
+  }
+}
--- a/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
@@ -1,0 +1,314 @@
+import {
+  ExpectedVersionConflictError,
+  STREAM_DOES_NOT_EXIST,
+  assertExpectedVersionMatchesCurrent,
+  type EventStore,
+  type Event,
+  type AggregateStreamOptions,
+  type AggregateStreamResult,
+  type AppendToStreamOptions,
+  type AppendToStreamResult,
+  type ReadStreamOptions,
+  type ReadStreamResult,
+  type ReadEventMetadata,
+  type ExpectedStreamVersion,
+} from '@event-driven-io/emmett';
+import mongoose from 'mongoose';
+
+export const MongoDBEventStoreDefaultStreamVersion = -1;
+
+export type StreamType = string;
+export type StreamName<T extends StreamType = StreamType> = `${T}:${string}`;
+
+export type StreamNameParts<T extends StreamType = StreamType> = {
+  streamType: T;
+  entityId: string;
+};
+
+export type StreamToProject<EventType extends Event> = {
+  streamName: StreamName;
+  streamType: StreamType;
+  entityId: string;
+  streamVersion: number;
+  events: EventType[];
+};
+
+export interface EventStream {
+  _id: mongoose.ObjectId;
+  streamName: string;
+  events: Array<{
+    _id: mongoose.ObjectId;
+    type: string;
+    data: string;
+    metadata: string;
+  }>;
+  createdAt: Date;
+  updatedAt: Date;
+}
+
+const eventStreamSchema = new mongoose.Schema(
+  {
+    streamName: {
+      type: String,
+      required: true,
+    },
+    events: {
+      type: [
+        {
+          type: {
+            type: String,
+            required: true,
+            enum: [],
+          },
+          data: {
+            type: String,
+            required: true,
+          },
+          metadata: {
+            type: String,
+            required: true,
+          },
+        },
+      ],
+      required: true,
+      default: [],
+    },
+  },
+  { timestamps: true },
+);
+eventStreamSchema.index({ streamName: 1 }, { unique: true });
+
+class EventStoreClass implements EventStore<number> {
+  private readonly Model: mongoose.Model<EventStream>;
+  constructor(modelName: string) {
+    this.Model =
+      mongoose.models[modelName] ??
+      mongoose.model<EventStream>(modelName, eventStreamSchema);
+  }
+
+  async readStream<EventType extends Event>(
+    streamName: StreamName,
+    options?: ReadStreamOptions<number>,
+  ): Promise<Exclude<ReadStreamResult<EventType, number>, null>> {
+    const expectedStreamVersion = options?.expectedStreamVersion;
+    const stream = await this.Model.findOne({
+      streamName: { $eq: streamName },
+    });
+
+    if (!stream) {
+      return {
+        events: [],
+        currentStreamVersion: MongoDBEventStoreDefaultStreamVersion,
+        streamExists: false,
+      };
+    }
+
+    assertExpectedVersionMatchesCurrent(
+      stream.events.length,
+      expectedStreamVersion,
+      MongoDBEventStoreDefaultStreamVersion,
+    );
+
+    const formattedEvents = stream.events
+      .slice(0, maxEventIndex(expectedStreamVersion))
+      .map(this.parseEvent<EventType>(streamName));
+
+    return {
+      events: formattedEvents,
+      currentStreamVersion: stream.events.length,
+      streamExists: true,
+    };
+  }
+
+  async aggregateStream<State, EventType extends Event>(
+    streamName: StreamName,
+    options: AggregateStreamOptions<State, EventType, number>,
+  ): Promise<AggregateStreamResult<State, number>> {
+    const stream = await this.readStream<EventType>(streamName, options?.read);
+    const state = stream.events.reduce(options.evolve, options.initialState());
+    return {
+      state,
+      currentStreamVersion: stream.currentStreamVersion,
+      streamExists: stream.streamExists,
+    };
+  }
+
+  async appendToStream<EventType extends Event>(
+    streamName: StreamName,
+    events: EventType[],
+    options?: AppendToStreamOptions<number> & {
+      /**
+       * This will be ran after a the events have been successfully appended to
+       * the stream. `appendToStream` will return after the project is completed.
+       */
+      project?: (stream: StreamToProject<EventType>) => void | Promise<void>;
+      /**
+       * Same as `options.project` but this will run asynchronously.
+       */
+      projectAsync?: (
+        stream: StreamToProject<EventType>,
+      ) => void | Promise<void>;
+    },
+  ): Promise<AppendToStreamResult<number>> {
+    const eventCreateInputs = events.map(this.stringifyEvent);
+
+    let stream = await this.Model.findOne({
+      streamName: { $eq: streamName },
+    }).lean();
+    let createdNewStream = false;
+
+    if (!stream) {
+      // @ts-expect-error
+      stream = await this.Model.create({
+        streamName,
+        events: [],
+      }).then((d) => d.toObject());
+      createdNewStream = true;
+    }
+
+    // NOTE: should never happen, the `create` call will throw an error if it fails
+    if (!stream) throw new Error('Failed to create stream');
+
+    assertExpectedVersionMatchesCurrent(
+      stream.events.length,
+      options?.expectedStreamVersion,
+      MongoDBEventStoreDefaultStreamVersion,
+    );
+
+    const updatedStream = await this.Model.findOneAndUpdate(
+      {
+        streamName: { $eq: streamName },
+        events: { $size: stream.events.length },
+      },
+      { $push: { events: { $each: eventCreateInputs } } },
+      { new: true },
+    ).lean();
+
+    if (!updatedStream) {
+      const currentStream = await this.Model.findOne(
+        { streamName: { $eq: streamName } },
+        { events: true },
+      ).lean();
+      throw new ExpectedVersionConflictError(
+        currentStream?.events.length ?? -1,
+        stream.events.length,
+      );
+    }
+
+    const { streamType, entityId } = fromStreamName(streamName);
+    if (options?.project) {
+      await options.project({
+        streamName,
+        streamType,
+        entityId,
+        streamVersion: updatedStream.events.length,
+        events: updatedStream.events.map(this.parseEvent(streamName)),
+      });
+    }
+
+    if (options?.projectAsync) {
+      options.projectAsync({
+        streamName,
+        streamType,
+        entityId,
+        streamVersion: updatedStream.events.length,
+        events: updatedStream.events.map(this.parseEvent(streamName)),
+      });
+    }
+
+    return {
+      nextExpectedStreamVersion: updatedStream.events.length,
+      createdNewStream,
+    };
+  }
+
+  /**
+   * Transforms the `event` from the saved format into the usable object
+   * at runtime. This function may be altered later to match `stringifyEvent`.
+   */
+  private parseEvent<EventType extends Event>(streamName: StreamName) {
+    return (
+      event: EventStream['events'][number],
+      index?: number,
+    ): EventType => {
+      const metadata = {
+        ...JSON.parse(event.metadata),
+        eventId: event._id,
+        streamName,
+        streamPosition: BigInt(index ?? 0),
+      } satisfies ReadEventMetadata;
+      return {
+        __brand: 'Event',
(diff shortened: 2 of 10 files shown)


alex-laycalvert:
Updated to remove `mongoose` in favor of the `mongodb` nodejs driver.

I'll leave some comments on the PR about certain decisions I made here.

src/packages/emmett-mongodb/package.json
```diff
@@ -0,0 +1,58 @@
...
+    }
+  },
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "dependencies": {
+    "mongodb": "^6.10.0"
```
alex-laycalvert, line 50: Might need to be a peer dep?

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,315 @@
...
+      index?: number,
+    ): ReadEvent<EventType, ReadEventMetadata> => {
+      const metadata = {
+        ...JSON.parse(event.metadata),
+        eventId: event._id,
+        streamName,
+        streamPosition: BigInt(index ?? 0),
+      } satisfies ReadEventMetadata;
+
+      // TODO: is this correct?
```
alex-laycalvert, line 230: I _think_ this is ok, TS just doesn't like it.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,315 @@
...
+      metadata: JSON.stringify(
+        event.metadata ?? {
+          now: new Date(),
+        },
+      ),
+    };
+  }
+}
+
+export const getMongoDBEventStore = async (
```
alex-laycalvert, line 261: I know the other event stores are sync here, but I want to add the `streamName` unique index which is an async call.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,315 @@
...
+       * These will be ran after a the events have been successfully appended to
+       * the stream. `appendToStream` will return after every projection is completed.
+       */
+      projections?: Array<
+        (stream: StreamToProject<EventType>) => void | Promise<void>
+      >;
+      /**
+       * Same as `options.projections` but this will run asynchronously.
+       */
+      asyncProjections?: Array<
```
alex-laycalvert, line 125: I think I've seen some other ways to handle registering projections in your talks so let me know if this isn't _the way_.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,315 @@
...
+      throw new ExpectedVersionConflictError(
+        currentStream?.events.length ?? -1,
+        stream.events.length,
+      );
+    }
+
+    const { streamType, entityId } = fromStreamName(streamName);
+
+    if (options?.projections) {
+      await Promise.all(
```
alex-laycalvert, line 183: Executed in no particular order, I don't think there would be a need to have ordered projections.

alex-laycalvert: Or at least not a need that couldn't be handled separately as an edge case by the dev

alex-laycalvert:
Been thinking about a global log position and I'm not sure if it's going to be possible, or at least simple to implement with mongodb.

Please correct me if I'm wrong here but it would need to be something like:

```typescript
// 1. Probably uising an aggregation, retrieve all events, order by global position, retrieve last one.
const lastEvent = collection.aggregate(...)

// 2. Using `lastEvent`, increment the global log position for each event we want to append
const eventInputs = [...]

// 3. Insert events into stream
collection.findOneAndUpdate(...)
```

But with the above impl, I don't know if we can guarantee that other events would have been made between step `2` and `3` creating a race condition.

We could use a global counter but I think it's susceptible to the same issues. Looking into possible locks on documents that we can have to make this simpler

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,315 @@
...
+}
+export type EventStreamEvent = EventStream['events'][number];
+
+export interface MongoDBConnectionOptions {
+  connectionString: string;
+  database: string;
+  collection?: string;
+}
+
+class EventStoreClass implements EventStore<number> {
```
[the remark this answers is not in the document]
alex-laycalvert: _outdated_

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,283 @@
...
+    this.collection = collection;
+  }
+
+  async readStream<EventType extends Event>(
+    streamName: StreamName,
+    options?: ReadStreamOptions<number>,
+  ): Promise<Exclude<ReadStreamResult<EventType, number>, null>> {
+    const expectedStreamVersion = options?.expectedStreamVersion;
+
+    // @ts-expect-error
```
alex-laycalvert, line 67: The only two uses of `@ts-expect-error` here are because the `collection` has to have a specific type but the type of each event stream will differ depending on the specific stream.

This is just to satisfy that the stream returned has the expected TS types.

alex-laycalvert: We could probably get away with this by adjusting the `findOne` and `findOneAndUpdate` type defs to allow a generic argument for the specific type with some `declare module ...` magic, but I'm afraid how that will affect people using this library and also using `mongodb`.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  async appendToStream<EventType extends Event>(
+    streamName: StreamName,
+    events: EventType[],
+    options?: AppendToStreamOptions<number> & {
+      /**
+       * These will be ran after a the events have been successfully appended to
+       * the stream. `appendToStream` will return after every projection is completed.
+       */
+      projections?: Array<Projection<EventType>>;
```
oskardudycz, line 119: **QUESTION:** @alex-laycalvert, could you expand on why you have the projections on the append level instead of the event store options (so as we have in PostgreSQL)?

alex-laycalvert: Let me look at the PostgreSQL example a bit more, there wasn't any specific reason I did this.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,272 @@
...
+}
+
+/**
+ * Accepts a `streamType` (the type/category of the event stream) and an `entityId`
+ * (the individual entity/object or aggregate ID) and combines them to a singular
+ * `streamName` which can be used in `EventStore`.
+ */
+export function toStreamName<T extends StreamType>(
+  streamType: T,
+  entityId: string,
```
oskardudycz, line 255: **SUGGESTION:** A bit nitpicking, but I'd prefer to call it `streamId`
```suggestion
  streamId: string,
```

oskardudycz: Note for me: Something for the future is to align this parsing throughout event stores (not this PR).

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,272 @@
...
+  type ReadStreamOptions,
+  type ReadStreamResult,
+  type ReadEvent,
+  type ReadEventMetadata,
+  type ExpectedStreamVersion,
+} from '@event-driven-io/emmett';
+import { type Collection, MongoClient, type WithId } from 'mongodb';
+import { v4 as uuid } from 'uuid';
+
+export const MongoDBEventStoreDefaultStreamVersion = -1;
```
oskardudycz, line 20: **SUGGESTION:** If there's no particular reason to use `-1`, I'd prefer to use `0` as the default version (so just like in PostgreSQL). That's more aligned with the numbering that databases have.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,272 @@
...
+export const MongoDBDefaultCollectionName = 'eventstreams';
```
oskardudycz, line 21: **SUGGESTION:** Something to do for the future would be to check if we should use a single collection for all streams, or break it down by stream type. It can stay like that, but later on some performance and load test would be needed to ensure that won't slow down the performance.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,272 @@
...
+export interface MongoDBConnectionOptions {
+  connectionString: string;
+  database: string;
+  collection?: string;
+}
+
+class EventStoreClass implements EventStore<number> {
+  private readonly collection: Collection<EventStream<Event>>;
+
+  constructor(collection: typeof this.collection) {
```
oskardudycz, line 61: **QUESTION:** Do we assume for now that we will perform all operations on the single collection side? @alex-laycalvert did you think already about transactions or sessions handling? This could be useful for e.g. projections to other collections.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,315 @@
...
+      });
+      stream = await this.collection.findOne({
+        _id: result.insertedId,
+      });
+      createdNewStream = true;
+    }
+
+    // TODO: better error here, should rarely happen if ever
+    // if another error was not thrown before this
+    if (!stream) throw new Error('Failed to create stream');
```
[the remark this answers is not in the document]
oskardudycz: **QUESTION:** @alex-laycalvert what's the expected behaviour in terms of eventual consistency, read-your-own-writes guarantees and retries? Technically, with Mongo, if we don't provide retries and we're not using write or read concerns, then reading right after the write may not show the same document, which can cause issues. I described those issues longer in https://event-driven.io/en/dealing_with_eventual_consistency_and_idempotency_in_mongodb_projections/

Did you already think about it? I think that we'll need to have a logic described as in the article in the end. (this can come in the follow up implementation tho).

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,272 @@
...
+
+    // @ts-expect-error The actual `EventType` is different across each stream document,
+    // but the collection was instantiated as being `EventStream<Event>`. Unlike `findOne`,
+    // `findOneAndUpdate` does not allow a generic to override what the return type is.
+    const updatedStream: WithId<EventStream<EventType>> | null =
+      await this.collection.findOneAndUpdate(
+        {
+          streamName: { $eq: streamName },
+          events: { $size: stream.events.length },
+          updatedAt: new Date(),
```
oskardudycz, line 175: **CONCERN:** Why do we have this date condition here?

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: I started this with the assumption that transactions aren't possible since my project doesn't use any replica sets. I'm definitely open to the idea of multiple collections based on stream type.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -0,0 +1,272 @@
...
+    );
+
+    // @ts-expect-error The actual `EventType` is different across each stream document,
+    // but the collection was instantiated as being `EventStream<Event>`. Unlike `findOne`,
+    // `findOneAndUpdate` does not allow a generic to override what the return type is.
+    const updatedStream: WithId<EventStream<EventType>> | null =
+      await this.collection.findOneAndUpdate(
+        {
+          streamName: { $eq: streamName },
+          events: { $size: stream.events.length },
```
oskardudycz, line 174: **FIX:** We should do optimistic concurrency update based on stream position (and probably also add it to an index, as we'll be searching on it). This will be faster, plus it'll enable scenarios like e.g. archiving older events in the future. Then after archiving, you could technically end up with the stream that doesn't have events, but the position is different than zero.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: Is it outdated?

alex-laycalvert: Yes

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: I think that we'll need to find a good way for that. Either using Mongo change streams to trigger those projections or having some sort collection with the outboxed messages. Have you thought about it already?

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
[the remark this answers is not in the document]
alex-laycalvert: I have not thought about it, I'll definitely take a look at the article.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: I think that's safer to run them sequentially as default and make it configurable to run them in parallel. As you wrote, the general practice is that projections should not depend on each other, but sometimes people does that, so following the least-surprise principle, I'd suggest as in the first sentence.

oskardudycz:
> Been thinking about a global log position and I'm not sure if it's going to be possible, or at least simple to implement with mongodb.
> 
> Please correct me if I'm wrong here but it would need to be something like:
> 
> ```ts
> // 1. Probably uising an aggregation, retrieve all events, order by global position, retrieve last one.
> const lastEvent = collection.aggregate(...)
> 
> // 2. Using `lastEvent`, increment the global log position for each event we want to append
> const eventInputs = [...]
> 
> // 3. Insert events into stream
> collection.findOneAndUpdate(...)
> ```
> 
> But with the above impl, I don't know if we can guarantee that other events would have been made between step `2` and `3` creating a race condition.
> 
> We could use a global counter but I think it's susceptible to the same issues. Looking into possible locks on documents that we can have to make this simpler

Definitely, for MongoDB we won't be providing the global ordering guarantee, as this will result in half-assed performance and unscalable solution.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: I'll be working on moving the projection registration to be options on the event store instead of per - append call so I'll probably follow the postgres example for this as well.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: I think that I'd then just update it and keep in metadata.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: Technically we should define our collection type. The set of metadata is known, and the events array type can be just default Event without defaulted generic parameters.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: Oop, actually, this was a typo.

I _meant_ to make the `updatedAt` as part of the update, not as part of the condition. My bad.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: I think that it's a safe choice to provide the initial capabilities without those additional features. We might still want to provide them through append options, but this can happen in a separate PR.

src/packages/emmett-mongodb/package.json
oskardudycz: Yup :+1:

oskardudycz:
@alex-laycalvert could you add the basic set of tests in the tests projects (as other stores): https://github.com/event-driven-io/emmett/tree/main/src/packages/emmett-tests/src/eventStore? After that, I'll be fine with merging it. We can tackle issues that we were discussing in separate PRs so as not to do all at once.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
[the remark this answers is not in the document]
oskardudycz: @alex-laycalvert ok, cool, I'm curious your thoughts on that. Still, that's something that we can/should do in a dedicated PR.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: I can keep these timestamps if we you think they'd be useful but really this info can be found from the first/last events too so probably just redundant.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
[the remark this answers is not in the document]
oskardudycz: Ah,and @alex-laycalvert the code I described in article, you can find here: 
- https://github.com/oskardudycz/EventSourcing.NodeJS/blob/0a08b4237b573747df7bacbf9ee195c4e8970edf/samples/webinar/src/core/mongoDB.ts#L102
- https://github.com/oskardudycz/EventSourcing.NodeJS/blob/0a08b4237b573747df7bacbf9ee195c4e8970edf/samples/webinar/src/shoppingCarts/shoppingCartDetails.ts#L104

Feel free to use it :)

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: That's why I was surprised, and why we need automated test 😅👌

oskardudycz:
> Sure thing @oskardudycz I'll clean up the thing about the `updatedAt` date condition (originally a typo), then add the tests.

Thanks a lot!

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: > I can keep these timestamps if we you think they'd be useful but really this info can be found from the first/last events too so probably just redundant.

I'm fine with having that also on the stream level. It's not taking much additional data, and we can use Mongo `project` not to read it, but just update it.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: Playing around with multiple collections separated by stream type, should we rename the `streamName` property on `EventStream` to just be the `streamId` instead of the fully formatted name including the stream type? Figured if each collection is the type, we can just use the streamId itself for each doc

alex-laycalvert:
Hey @oskardudycz , let me know if there's anything else you think we should add to this initial implementation before it can get merged, thanks!

oskardudycz:
@alex-laycalvert, apologies for the delay. I've been constantly traveling for the last two weeks, and that got me a bit off schedule. I'm pulling that in. I'll try to write all the notes and issues based on the discussion in this PR about the next steps (like e.g. projections, retries, etc.). And tag you in there.

Would you like me to release the package as it is?

oskardudycz:
@alex-laycalvert and again, thank you a lot for your work, and contributing it back ❤️

PR #127, opened 2024-11-12
[MongoDB Event Store] Projection Registration & Helper for single stream projections

alex-laycalvert:
- Changed the registration of projections to be event store wide options instead of per `appendToStream` call
- The "short info" read models are now stored directly on the event stream instead of a separate collection.
- Removed unused code
- Changed stream version of mongodb event store to `BigInt`.

The `streamPosition` is stored as a `string` value since MongoDB doesn't directly support storing `BigInt` objects from JS.

### DIFF
--- a/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
@@ -102,7 +102,11 @@ void describe('MongoDBEventStore', () => {
     assertDeepEqual(stream.projections[SHOPPING_CART_PROJECTION_NAME], {
       productItemsCount: 20,
       totalAmount: 54,
-      _metadata: { name: SHOPPING_CART_PROJECTION_NAME, streamPosition: 3n },
+      _metadata: {
+        name: SHOPPING_CART_PROJECTION_NAME,
+        streamPosition: 3n,
+        schemaVersion: 1,
+      },
     });
   });
 
--- a/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
@@ -124,13 +124,17 @@ void describe('EventStoreDBEventStore', () => {
       { expectedStreamVersion: STREAM_DOES_NOT_EXIST },
     );
 
-    const expectedStreamVersion = 2;
+    const expectedStreamVersion = 3n;
+    const expectedNumEvents = 2;
     const stream = await eventStore.readStream<ShoppingCartEvent>(streamName, {
-      expectedStreamVersion: BigInt(expectedStreamVersion),
+      from: 0n,
+      to: BigInt(expectedNumEvents),
+      expectedStreamVersion,
     });
 
     assertTrue(stream.streamExists);
-    assertEqual(expectedStreamVersion, stream.events.length);
+    assertEqual(expectedStreamVersion, stream.currentStreamVersion);
+    assertEqual(expectedNumEvents, stream.events.length);
   });
 });
 
--- a/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
@@ -3,13 +3,14 @@ import {
   assertEqual,
   assertIsNotNull,
   assertTrue,
+  projections,
   STREAM_DOES_NOT_EXIST,
 } from '@event-driven-io/emmett';
 import {
   MongoDBContainer,
   type StartedMongoDBContainer,
 } from '@testcontainers/mongodb';
-import { type Collection, MongoClient } from 'mongodb';
+import { MongoClient, type Collection } from 'mongodb';
 import { after, before, describe, it } from 'node:test';
 import { v4 as uuid } from 'uuid';
 import {
@@ -21,15 +22,15 @@ import {
 import {
   getMongoDBEventStore,
   toStreamName,
-  mongoDBInlineProjection,
   type EventStream,
   type MongoDBEventStore,
 } from './mongoDBEventStore';
+import { mongoDBInlineProjection } from './projections';
 
 const DB_NAME = 'mongodbeventstore_testing';
 const SHOPPING_CART_PROJECTION_NAME = 'shoppingCartShortInfo';
 
-void describe('EventStoreDBEventStore', () => {
+void describe('MongoDBEventStore', () => {
   let mongodb: StartedMongoDBContainer;
   let eventStore: MongoDBEventStore;
   let client: MongoClient;
@@ -49,13 +50,13 @@ void describe('EventStoreDBEventStore', () => {
 
     eventStore = getMongoDBEventStore({
       collection,
-      projections: [
+      projections: projections.inline([
         mongoDBInlineProjection({
           name: SHOPPING_CART_PROJECTION_NAME,
           canHandle: ['ProductItemAdded', 'DiscountApplied'],
           evolve,
         }),
-      ],
+      ]),
     });
     return eventStore;
   });
@@ -92,12 +93,16 @@ void describe('EventStoreDBEventStore', () => {
       { expectedStreamVersion: STREAM_DOES_NOT_EXIST },
     );
 
-    const stream = await collection.findOne({ streamName });
+    const stream = await collection.findOne(
+      { streamName },
+      { useBigInt64: true },
+    );
     assertIsNotNull(stream);
     assertEqual('3', stream.metadata.streamPosition.toString());
-    assertDeepEqual(stream.projections[SHOPPING_CART_PROJECTION_NAME]?.short, {
+    assertDeepEqual(stream.projections[SHOPPING_CART_PROJECTION_NAME], {
       productItemsCount: 20,
       totalAmount: 54,
+      _metadata: { name: SHOPPING_CART_PROJECTION_NAME, streamPosition: 3n },
     });
   });
 
@@ -146,7 +151,7 @@ type ShoppingCartShortInfo = {
 const evolve = (
   document: ShoppingCartShortInfo | null,
   { type, data: event }: ProductItemAdded | DiscountApplied,
-): ShoppingCartShortInfo => {
+): ShoppingCartShortInfo | null => {
   document = document ?? { productItemsCount: 0, totalAmount: 0 };
 
   switch (type) {
--- a/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
@@ -1,9 +1,13 @@
-import { assertDeepEqual, assertIsNotNull } from '@event-driven-io/emmett';
+import {
+  assertDeepEqual,
+  assertEqual,
+  assertIsNotNull,
+} from '@event-driven-io/emmett';
 import {
   MongoDBContainer,
   type StartedMongoDBContainer,
 } from '@testcontainers/mongodb';
-import { MongoClient } from 'mongodb';
+import { type Collection, MongoClient } from 'mongodb';
 import { after, before, describe, it } from 'node:test';
 import { v4 as uuid } from 'uuid';
 import {
@@ -15,29 +19,41 @@ import {
 import {
   getMongoDBEventStore,
   toStreamName,
+  shortInfoProjection,
   type EventStream,
   type MongoDBEventStore,
 } from './mongoDBEventStore';
 
 const DB_NAME = 'mongodbeventstore_testing';
-const SHOPPING_CARD_INFO_COLLECTION_NAME = 'shoppingCartShortInfo';
 
 void describe('EventStoreDBEventStore', () => {
   let mongodb: StartedMongoDBContainer;
   let eventStore: MongoDBEventStore;
   let client: MongoClient;
+  let collection: Collection<EventStream>;
 
   before(async () => {
     mongodb = await new MongoDBContainer().start();
     client = new MongoClient(mongodb.getConnectionString(), {
       directConnection: true,
     });
+
     await client.connect();
     const db = client.db(DB_NAME);
-    const collection = db.collection<EventStream>(
+    collection = db.collection<EventStream>(
       'mongodbeventstore_testing_eventstreams',
     );
-    eventStore = getMongoDBEventStore(collection);
+
+    eventStore = getMongoDBEventStore({
+      collection,
+      projections: [
+        shortInfoProjection({
+          name: 'shoppingCartShortInfo',
+          canHandle: ['ProductItemAdded', 'DiscountApplied'],
+          evolve,
+        }),
+      ],
+    });
     return eventStore;
   });
 
@@ -58,15 +74,10 @@ void describe('EventStoreDBEventStore', () => {
     };
     const discount = 10;
     const shoppingCartId = uuid();
-
-    const shoppingCartShortInfo = client
-      .db()
-      .collection<
-        ShoppingCartShortInfo & { streamId: string; version: number }
-      >(SHOPPING_CARD_INFO_COLLECTION_NAME);
+    const streamName = toStreamName('shopping_cart', shoppingCartId);
 
     await eventStore.appendToStream<ShoppingCartEvent>(
-      toStreamName('shopping_cart', shoppingCartId),
+      streamName,
       [
         { type: 'ProductItemAdded', data: { productItem } },
         { type: 'ProductItemAdded', data: { productItem } },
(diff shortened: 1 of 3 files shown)


src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -27,51 +31,63 @@ export type StreamNameParts<T extends StreamType = StreamType> = {
...
-  createdAt: Date;
-  updatedAt: Date;
+  metadata: {
+    streamId: string;
+    streamType: StreamType;
+    streamPosition: bigint;
+    createdAt: Date;
+    updatedAt: Date;
+  };
+  projection: {
```
oskardudycz, line 47: **FIX:** I think that it'd be better to make projection either a list of projections or a record. People can have more than two, or have just one, or none.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -82,22 +98,23 @@ export class MongoDBEventStore implements EventStore<number> {
...
 
     assertExpectedVersionMatchesCurrent(
-      stream.events.length,
+      stream.metadata.streamPosition,
       expectedStreamVersion,
       MongoDBEventStoreDefaultStreamVersion,
     );
 
     return {
+      // TODO: remove `.slice`?
```
oskardudycz, line 107: **COMMENT:** Slice is needed when we support reading a part of stream, see, e.g. how PostgreSQL filter implementation: https://github.com/event-driven-io/emmett/blob/97519d0b561943a4a096f44cf95c4c8fd1393de8/src/packages/emmett-postgresql/src/eventStore/schema/readStream.ts#L40. Here we could either use project from mongo, to just read the subset of array, or filter it with slice.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -110,31 +127,38 @@ export class MongoDBEventStore implements EventStore<number> {
...
-      projections?: Array<Projection<EventType>>;
-    },
-  ): Promise<AppendToStreamResult<number>> {
-    let stream = await this.collection.findOne({
-      streamName: { $eq: streamName },
-    });
-    let currentStreamPosition = stream?.events.length ?? 0;
+    options?: AppendToStreamOptions,
+  ): Promise<AppendToStreamResult> {
+    let stream = streamPositionDeserializer(
```
oskardudycz, line 132: **FIX:** This could be made in a another PR, but when we're appending the stream, then we should not be reading it, we can use retries and checks on the expected version to ensure that we're not overriding the stream in the unexpected state.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -154,89 +178,154 @@ export class MongoDBEventStore implements EventStore<number> {
...
+  const eventStore = new MongoDBEventStore(options);
   return eventStore;
 };
 
-function executeProjections<EventType extends Event>(
-  params: StreamToProject<EventType>,
-  projections?: Array<Projection<EventType>>,
-) {
-  return Promise.all((projections ?? []).map((project) => project(params)));
+async function handleProjections<
```
oskardudycz, line 251: **FIX:** So we have here two types of projections:
1. Projections to the same document (let's call them _"inline"_).
2. Projections to external collection.

For the first type of projections, it'd be better to use `$set` and store them in the same update as appending events. Then you have the certainty that you won't have race condition and change will be atomic.

For the second part, what you did here should be fine, of course keeping in mind, that process may die in between and those other projections won't be updated, and that we need to apply retries, as mentioned in the first PR.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -264,3 +353,28 @@ export function fromStreamName<T extends StreamType>(
...
+
+/**
+ * Converts the `stream.metadata.streamPosition` of the given `stream`
+ * to a `string` value to be stored in MongoDB
+ */
+function streamPositionSerializer<Stream extends EventStream>(
+  stream: Stream,
+): Stream {
+  // @ts-expect-error serializing as a `string`
+  stream.metadata.streamPosition = stream.metadata.streamPosition.toString();
```
oskardudycz, line 365: **CONCERN:** Wouldn't be enough to use [useBigInt64 option](https://mongodb.github.io/node-mongodb-native/5.2/interfaces/MongoClientOptions.html#useBigInt64)? Read also more [here](https://www.mongodb.com/community/forums/t/mongodb-nodejs-driver-5-1-0-released/214717).

oskardudycz (changes requested):
@alex-laycalvert thanks for sending the PR! Could you check my comments? :)

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: Yep I've been thinking about that as I was working on this

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: I didn't know about that option, I may have been seeing some outdated forums saying string was best

oskardudycz: To be fair, I also found that option today (tho, I knew that BSON on its own can handle BigInts).  😅

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: I think I'll do this at the DB level since it supports it. Would `currentStreamPosition` remain as the current version of the whole stream or should it reflect what was returned based on `expectedStreamVersion`

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: I had the same thought for inline projections, but got a little stuck on aggregating the events that are being appended and getting the correct initial state. Should inline projections _return_ the state to be added to the document? That way the appendToStream func has something to add to the update body or is there a better way to handle that?

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: How would you recommend setting the `streamPosition` at the event metadata level when we go to append the events? Or is it not possible to store that without reading the stream first, and can just be encapsulated by the index in the array?

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: I think that there are multiple options to handle inline projections:
- return the updated projections object,
- return a Map of updated projections,
- return the updated syntax of MongoDB.

I think that we can start with what's the easiest to support. I would vote to do the following when the projections are registered:
1. Read the projections subdocument (so it contains all inline projections).
2. Run handle methods on top of it. The single inline projection registration would need to contain key of the projection in the projection subdocument. This key could be optional and fall back to "default" if not provided, as my guess is that most of the people would just have one document.
3. Update it in the same change using `$set` operator.

@alex-laycalvert, thoughts?

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: Good question! I think I'd still return the current version of the whole stream, as this is meant to be used later in the optimistic concurrency check. Plus, the name suggests that it should be the current stream position, so I think that's worth following the least surprise principle. If someone wants to read the last position of read events, they'll be able to do it from event metadata. Maybe in the future, it'd be worth adding additional property if it was used more often.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: That makes sense, I think the `name` that we register the projection with should suffice as a key, stored in `details.name` on the projection. We could also add a separate key property.

I'll play around with this

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: @alex-laycalvert Yes, it's possible; if you don't provide append options or choose to use no concurrency check, then the event should be just appended and the version auto-incremented. If autoincrement doesn't work for bigints in MongoDB, then we can use number for stream position.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: @alex-laycalvert Sounds good. 👌

It would be also good to double check if our structure is good for indexing strategies. I'm not sure what plays well in Mongo :+1:

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: Does MongoDB natively support an auto incrementing field or would this be in the form of a separate document as a `counter`?

oskardudycz: It does, by using $inc operator https://www.mongodb.com/docs/manual/reference/operator/update/inc/.

At least, that’s how I did it so far.

oskardudycz:
@alex-laycalvert thank you for applying suggestions. I’ll go through them tomorrow 👌

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
```diff
@@ -75,35 +89,48 @@ void describe('EventStoreDBEventStore', () => {
...
+      ],
+      { expectedStreamVersion: STREAM_DOES_NOT_EXIST },
+    );
 
-    assertDeepEqual(docWithoutId, {
-      streamId: shoppingCartId,
-      version: 3,
-      productItemsCount: 20,
-      totalAmount: 54,
+    const expectedStreamVersion = 2;
```
oskardudycz, line 127: **QUESTION:** shouldn't we have here version number 3? As we're adding 3 events. I think that we discussed to start counting stream position from 1.

alex-laycalvert: Correct I forgot to adjust this

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -27,51 +30,100 @@ export type StreamNameParts<T extends StreamType = StreamType> = {
   streamId: string;
 };
 
-export type StreamToProject<EventType extends Event> = {
+export interface EventStreamProjection<
+  ShortInfoType extends DefaultRecord = DefaultRecord,
+> {
+  details: { name?: string };
+  short: ShortInfoType | null;
```
oskardudycz, line 37: **QUESTION:** Could you explain me why we still have short? property here? Is it needed? I think that we could just keep the regular map of object inside the projections property of stream. Could you explain why do we need this additional nesting?

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
alex-laycalvert: Actually I'm only using this to confirm we only receive that many events from the stream if passed as a filter

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -27,51 +30,100 @@ export type StreamNameParts<T extends StreamType = StreamType> = {
...
 
-export interface EventStream<EventType extends Event = Event> {
-  streamName: string;
-  events: Array<ReadEvent<EventType, ReadEventMetadata>>;
-  createdAt: Date;
-  updatedAt: Date;
-}
-export type EventStreamEvent<EventType extends Event = Event> =
-  EventStream<EventType>['events'][number];
+type MongoDBInlineProjection<
```
oskardudycz, line 65: **CONCERN:** @alex-laycalvert is there a reason why you not extend the ProjectionDefinition from the regular Emmett package? See https://github.com/event-driven-io/emmett/blob/97519d0b561943a4a096f44cf95c4c8fd1393de8/src/packages/emmett/src/projections/index.ts#L30

And how PostgreSQL projection does it: https://github.com/event-driven-io/emmett/blob/97519d0b561943a4a096f44cf95c4c8fd1393de8/src/packages/emmett-postgresql/src/eventStore/projections/index.ts#L35.

Also, best if projection definition is distinct from the registration (inline, async) so someone could change whether it's run inline or async without changing the projection definition.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -110,31 +163,34 @@ export class MongoDBEventStore implements EventStore<number> {
...
-      projections?: Array<Projection<EventType>>;
-    },
-  ): Promise<AppendToStreamResult<number>> {
-    let stream = await this.collection.findOne({
-      streamName: { $eq: streamName },
-    });
-    let currentStreamPosition = stream?.events.length ?? 0;
+    options?: AppendToStreamOptions,
+  ): Promise<AppendToStreamResult> {
+    let stream = await this.collection.findOne(
```
oskardudycz, line 168: **CONCERN:** Is it intentional, that kept here the reading the document? We discussed in the other thread, not to do it unles there are inline projections registered.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -154,89 +210,159 @@ export class MongoDBEventStore implements EventStore<number> {
...
+    const updates: UpdateFilter<EventStream> = {
+      $push: { events: { $each: eventCreateInputs } },
+      $set: {
+        'metadata.updatedAt': new Date(),
+        'metadata.streamPosition':
+          stream.metadata.streamPosition + BigInt(events.length),
+      },
+    };
+
+    if (this.asyncProjections) {
```
oskardudycz, line 231: **CONCERN:** Async projections should be handled after events are appended. Without that, you may update async projections and not have event appended. which will cause state inconsistency.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -154,89 +210,159 @@ export class MongoDBEventStore implements EventStore<number> {
...
-        { returnDocument: 'after' },
+        updates,
+        { returnDocument: 'after', useBigInt64: true },
       );
 
     if (!updatedStream) {
-      const currentStream = await this.collection.findOne({
-        streamName: { $eq: streamName },
-      });
+      const currentStream = await this.collection.findOne(
```
oskardudycz, line 268: **CONCERN:** I'd prefer not to have the actual position here in error and not need to read the whole document again. Especially that's not a precise information, as someone else could make the update in the  meantime.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -154,89 +210,159 @@ export class MongoDBEventStore implements EventStore<number> {
...
+  canHandle: CanHandle<EventType>;
+  evolve: (state: ShortInfoType, event: EventType) => ShortInfoType;
+}): MongoDBInlineProjection {
+  return {
+    type: 'inline',
+    projection: {
+      name: options.name,
+      canHandle: options.canHandle,
+      handle: async (events, { streamName, collection }) => {
+        const stream = await collection.findOne({ streamName });
```
oskardudycz, line 333: We should not be loading anything additionally in inline projection, as we already should have already this data in the stream state.

oskardudycz:
@alex-laycalvert, I added some follow up comments/concerns. If they're unclear feel free to ping me here, on Discord. I'm also happy to add commits on my own to apply them if that would be easier for you.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.e2e.spec.ts
alex-laycalvert: Oh I see now, that makes more sense.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: Removed

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: Moved to after the events are appended

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: Think I might need your help on this for inline projections. I can't seem to figure out how we can have the projection handler be passed the stream state if we run the inline projections before the update so that we can have a single update instead of multiple when appending to the stream.

It feels like there's kind of a dependency circle between the handler needing to know the initial state, either current short projection on doc (requires a read), or the inline projections would be passed all events on the stream and not just the ones about to append, but that would need to be done after the stream update meaining a second update required for the projection handler if that makes sense.

oskardudycz: I'll adjust the type definition in https://github.com/event-driven-io/emmett/pull/132. I think that the best way will be to return inline projections together with the event stream and allow the reading result to be passed to append as it was (e.g. ` { expectedStreamState: ReadResult }` then we can read that from read result and assume that's the current projection state. Together with optimistic concurrency and atomic update, it should be sufficient.

For that, I'll need to adjust the command handler and event store definition to allow the extended event stream result to be taken (specific for event store type).

alex-laycalvert:
Fixed merge conflicts, having some build issues locally for some reason, I'll wait to see if it's affecting the GH checks as well.

oskardudycz (approved):

PR #134, opened 2024-12-05
[MongoDB Event Store]: Fix metadata properties / export `emmett-mongodb/eventStore/projections`

alex-laycalvert:
- Included `eventStore/projections` export in the main mongodb event store module. Was left out so the helpers in `projections` were not useable
- Several metadata properties were not being created on stream creation in the event store. Added these in in `appendToStream` with `$setOnInsert`
- Added test to ensure metadata properties are set on stream creation


oskardudycz (approved):
Thank you @alex-laycalvert, I somehow missed it!

PR #139, opened 2024-12-07
[MongoDB EventStore] Add query functions for inline projections

### LINKED ISSUE #137: [MongoDB EventStore] Inline Projection Helper for reading from read models
From @oskardudycz comment in discord, I think a helper reading from projections would be very helpful.

```typescript
eventStore.projections.inline.findOne(streamType, {
    // ...
})

// or

eventStore.readModels.findOne(streamType, {
    // ...
})
```

`streamType` will be required since we're splitting up collections via type in #136 . 

I think we can essentially make the second argument mimic filter queries of mongodb itself, but unless we only allow reading from a pre-set projection name and instead allow searching of all projections on all streams, I think we need to do something like this:

```javascript
$expr: {
    $gt: [
      {
        $size: {
          $filter: {
            input: { $objectToArray: "$docs" }, // Convert "docs" to an array of key-value pairs
            as: "doc",
            cond: { $eq: ["$$doc.v.someValue", "desiredValue"] } // Check the nested field
          }
        }
      },
      0
    ]
  }
```

I _assume_ as long as the number of projections on each is small, then this doesn't come as too much of a performance cost but I also guess that it's still slow than only allow searching a single projection on each stream like `projections.READ_MODEL.someValue: property`.

I'm not an expert here with mongo optimizations so maybe there's a simpler way.

alex-laycalvert:
## Description

- Add support for `projections.inline.findOne(...)` for inline projections
- Add support for `projections.inline.find(...)` for inline projections
- Inline proj. queries mimic mongoDB queries in syntax but don't require appending `projection.${projName}.value` and instead can just be `value: { ... }` (proper prefixes are added to each key that's not a mongoDB operator like `$in` or the specific `streamName` property that doesn't need a prefix.
- `streamId` is added to each projection `_metadata`. Useful for identifying specific entity when all you have is the projection
- E2E tests for new behavior
- `streamType` is a required param for each inline proj. query, `projectionName` is optional and if not provided will use `_default` (default name for adding projections without an explicit name param)
- Adds a unique index on each stream type collection
- Adds basic pagination/sorting options for `projections.inline.find`

Example usage so far:

```typescript
const eventStore = getMongoDBEventStore({
    // ... other options
    projections: [mongoDBInlineProjection({ name: 'ShoppingCart' })],
})

// ... append to stream events of type 'shoppingCard'

const shoppingCart = await eventStore.projections.inline.findOne(
    'shoppingCart', // stream type
    'ShoppingCart', // projection name
    
    // the mongodb query
    {
        productItemsCount: { $eq: 20 },
        totalAmount: { $gt: 10 },
        '_metadata.schemaVersion': 1
    },
)

// shoppingCart is the inner projection data for the 'ShoppingCart' projection name
```

## Pre-Requisites

Pulled initial branch directly from this PR

- #136

## TODO

- [x] Pagination for `.find`. It _might_ be best to instead expose the "raw" mongo query for this function instead of evaluating it for the user, letting them do operations like `skip` or `limit`. Not sure how in depth we want to go but pagination I think should be added.

## Resolved Issues

- Fixes #137

### DIFF
--- a/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
@@ -17,8 +17,11 @@ import {
   type ReadStreamResult,
 } from '@event-driven-io/emmett';
 import {
+  Db,
+  MongoClient,
   type Collection,
   type Document,
+  type MongoClientOptions,
   type UpdateFilter,
   type WithId,
 } from 'mongodb';
@@ -37,6 +40,13 @@ export type StreamName<T extends StreamType = StreamType> = `${T}:${string}`;
 export type StreamNameParts<T extends StreamType = StreamType> = {
   streamType: T;
   streamId: string;
+};
+
+export type StreamCollectionName<T extends StreamType = StreamType> =
+  `emt:${T}`;
+
+export type StreamCollectionNameParts<T extends StreamType = StreamType> = {
+  streamType: T;
 };
 
 export type MongoDBReadModelMetadata = {
@@ -67,12 +77,6 @@ export interface EventStream<
   projections: Record<string, MongoDBReadModel>;
 }
 
-export interface MongoDBConnectionOptions {
-  connectionString: string;
-  database: string;
-  collection?: string;
-}
-
 export type MongoDBReadEventMetadata =
   ReadEventMetadataWithoutGlobalPosition<bigint>;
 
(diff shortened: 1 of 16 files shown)


alex-laycalvert:
@oskardudycz I didn't rebase this by accident so not sure if having your earlier commits would mess up the git history too much, I don't think it should be an issue though

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -378,6 +408,122 @@ class MongoDBEventStoreImplementation implements MongoDBEventStore {
     if (!this.isClosed) await this.client.connect();
     return this.client;
   };
+
+  private async findOneInlineProjection<
+    Doc extends Document,
+    Projection = MongoDBReadModel<Doc>,
+  >(
+    streamType: StreamType,
```
oskardudycz, line 416: **SUGGESTION:** I think that we could also allow passing stream name to find one, then we'd automatically allow selecting the read model by the ID without additional parameters. So we could do something like:

````ts
      await eventStore.projections.inline.findOne<ShoppingCartShortInfo>(
        streamName
      );
```

Thoughts?

Or we could introduce the conventional ID for projection name like we did for stream name. It could be streamType: streamId: projectionName. Then, we could allow passing stream type, stream name, or projection name in this first parameter. Thoughts?

oskardudycz:
> Also, I think while these functions might help we might also want to expose a prefix helper for generating queries. I know of some of my models that I'll need to do aggregations on and will need some more complex logic that we shouldn't expect to support here.
>
> Maybe export a DefaultProjectionName const for people to use to prefix their type names.

Yup, we could expose some helper functions that'd generate the filter, so people could use them in the regular collection usage.

Regarding paging and filtering, I think that's a good  idea, still, we can take it as a separate PR :)

> @oskardudycz I didn't rebase this by accident so not sure if having your earlier commits would mess up the git history too much, I don't think it should be an issue though

No worries, I'll then squash and merge changes :+1:

oskardudycz (approved):
I added suggestion, but even without applying it, I think that's good as it is. Nicely done! 🙂

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
alex-laycalvert: Yeah I like that. Right now I left that as an available option you could provide in the query itself, but I think for these stream level properties, maybe we make the first param an object that is either the `streamType` or can be `streamName` or optional `streamId`

alex-laycalvert:
```typescript
// querying over whole collection
projections.inline.findOne({ streamType }, ...)

// finding specific stream projection
projections.inline.findOne({ streamName }, ...)

// streamId also optional, maybe just for convenience whether you already
// have a formatted name or just the parts.
projections.inline.findOne({ streamType, streamId }, ...)
```

oskardudycz: Yes, an object also works; we can always make it either an object or a conventional ID; depending on someone's preferences, they can use it either way. The object will probably be the most accessible to folks.

src/packages/emmett-mongodb/src/eventStore/projections/index.ts
```diff
@@ -186,12 +192,13 @@ export const mongoDBInlineProjection = <
...
 
       const metadata: MongoDBReadModelMetadata = {
+        streamId,
         name: projectionName,
         schemaVersion,
         streamPosition: events[events.length - 1]!.metadata.streamPosition,
       };
 
-      updates.$set![`projections.${options.name}`] =
+      updates.$set![`projections.${projectionName}`] =
```
oskardudycz, line 201: @alex-laycalvert thanks for fixing my mishap, I noticed this bug today! 😅❤️

oskardudycz (approved):
Nice job @alex-laycalvert !

PR #147, opened 2024-12-13
`eventStore.projections.inline.count` Projection Query Helper

alex-laycalvert:
## Description

- Adds `eventStore.projections.inline.count` which uses the same query filters as `projecitons.inline.find` but returns the result of `countDocuments` on the projections


alex-laycalvert:
I'm not in a rush with this one, the models I'm using the projection queries for don't have a lot of docs right now so "manual" pagination isn't too bad. This can wait until another bigger release if needed.

oskardudycz (approved):
@alex-laycalvert lgtm :+1: Thanks!

PR #150, opened 2024-12-17
Made the MongoDB event store explicit on whether it can close MongoDBclient or not

### LINKED ISSUE #143: Wrap MongoDB setup to return explicit store with close method when it manages Mongo client
See more in [discussion with @alex-laycalvert](https://github.com/event-driven-io/emmett/pull/136#discussion_r1874507468).

oskardudycz:
So far, the MongoDB event store has always exposed the close method. That's redundant for the case when the client is passed through options.

Now, it'll only be exposed if the store is created based on the connection string.

Following the Boy Scout rule, I also added a general Closeable interface to facilitate uniform handling of such cases in the future.

Follow up to #136

Fixes #143

### DIFF
--- a/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.connection.e2e.spec.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.connection.e2e.spec.ts
@@ -48,6 +48,7 @@ void describe('MongoDBEventStore connection', () => {
 
   void it('connects using connection string', async () => {
     const eventStore = getMongoDBEventStore({
+      storage: { type: 'COLLECTION_PER_STREAM_TYPE' },
       connectionString: mongodb.getConnectionString(),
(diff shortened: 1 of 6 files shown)


src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.connection.e2e.spec.ts
```diff
@@ -121,37 +120,6 @@ void describe('MongoDBEventStore connection', () => {
     }
   });
 
-  void it('does not disconnect connected client', async () => {
```
oskardudycz, line 124: **Note:** It's nice that we can remove tests; it's no longer needed, as this use case won't even be allowed.

oskardudycz:
@alex-laycalvert, could you have a look when you have time? :slightly_smiling_face:

oskardudycz:
@alex-laycalvert, thank you for checking it!

PR #151, opened 2024-12-17
Added explicit storage type selection for MongoDB Event Store

### LINKED ISSUE #142: Add an explicit switch for MongoDB collection storage
Per [discussion with @alex-laycalvert](https://github.com/event-driven-io/emmett/pull/136#discussion_r1874507134). We should add a switch to options to select explicitly which type of storage the user should choose: single or multi-collection. We should also allow to provide own collection (and db) resolution strategy.

oskardudycz:
Exposed strongly typed storage options for MongoDB Event Store. Now they can be explicitly selected and mutually exclusive:
- `COLLECTION_PER_STREAM_TYPE` - default option, where each stream type has a dedicated storage. As we're keeping _inline_ projections in the same document as the stream, then that gives a better option for custom indexing and better storage usage.
- `SINGLE_COLLECTION` can be either custom or default to `emt:streams`. This can be a fair choice if you don't have a large number of streams or you don't need customisation per stream type.
- `CUSTOM` - you can define your own collection resolution based on the stream type. It might be that you need to route databases per tenant or handle some other more advanced stuff, so this is the choice for you.

Added the possibility of providing custom collection and database resolution.

Moved storage-related code to a dedicated file, as it got a bit too crowded in the MongoDBEventStore implementation.

Fixes #142

### DIFF
--- a/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.connection.e2e.spec.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.connection.e2e.spec.ts
@@ -1,30 +1,12 @@
-import {
-  assertIsNotNull,
-  assertNotEqual,
-  assertThrowsAsync,
-  STREAM_DOES_NOT_EXIST,
-} from '@event-driven-io/emmett';
+import { assertIsNotNull, assertThrowsAsync } from '@event-driven-io/emmett';
 import {
   MongoDBContainer,
   type StartedMongoDBContainer,
 } from '@testcontainers/mongodb';
 import { MongoClient, MongoNotConnectedError } from 'mongodb';
 import { after, before, describe, it } from 'node:test';
-import { v4 as uuid } from 'uuid';
-import {
-  getMongoDBEventStore,
-  MongoDBEventStoreDefaultStreamVersion,
-  toStreamCollectionName,
-  toStreamName,
-  type MongoDBEventStore,
-  type StreamType,
-} from '.';
-import {
-  type PricedProductItem,
-  type ShoppingCartEvent,
-} from '../testing/shoppingCart.domain';
-
-const streamType: StreamType = 'shopping_cart';
+import { getMongoDBEventStore, toStreamCollectionName } from '.';
+import { assertCanAppend, ShoppingCartStreamType } from '../testing';
 
 void describe('MongoDBEventStore connection', () => {
   let mongodb: StartedMongoDBContainer;
@@ -86,7 +68,7 @@ void describe('MongoDBEventStore connection', () => {
       // this should succeed as event store should call connect internally
       const stream = await client
         .db()
-        .collection(toStreamCollectionName(streamType))
+        .collection(toStreamCollectionName(ShoppingCartStreamType))
         .findOne();
       assertIsNotNull(stream);
     } finally {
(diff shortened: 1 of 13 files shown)


oskardudycz:
@alex-laycalvert, I need to add tests for the collection resolution-related code, but could you check if this works for you?

I was going back and forth on whether we should make the storage setting explicit or not. On the one hand, it's an important decision; on the other, many new users probably won't know which one is better, so we'll still need to document it. So, at this point, I decided to make the default choice implicit to collection per stream type to encourage people to use the default.

oskardudycz:
It sounds like I was a bit sloppy and the commit https://github.com/event-driven-io/emmett/pull/150/commits/cc43517129f116a7ac18b959cedc56afadfe023c got into #150 instead of here. I’ll revert it on main to keep a full diff here.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.connection.e2e.spec.ts
```diff
@@ -144,50 +111,10 @@ void describe('MongoDBEventStore connection', () => {
     }
   });
 
-  void it('sets up database and collection', async () => {
```
oskardudycz, line 147: **COMMENT:** Storage tests were moved to the specific file.

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.storage.e2e.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+): Promise<Collection<EventStream<ProductItemAdded | DiscountApplied>>> => {
+  const existingCollections = await db.collections();
+  const collection = existingCollections.find(
+    (c) => c.collectionName,
+    collectionName,
+  );
+
+  assertOk(collection);
+
+  const indexes = await collection.indexes();
```
oskardudycz, line 197: **COMMENT:** I'm testing indexes in e2e, as I want to ensure they're setup. Testing that in the unit tests would require too much mambo jumbo around mocking, which I don't like.

src/packages/emmett-mongodb/src/eventStore/storage/mongoDBEventStoreStorage.unit.spec.ts
```diff
@@ -0,0 +1,432 @@
...
+  let otherStreamTypeName: OtherTestStreamType;
+  let defaultDBName: string;
+  let getConnectedClient: () => Promise<MongoClient>;
+
+  beforeEach(() => {
+    testStreamTypeName = uuid();
+    otherStreamTypeName = uuid();
+    defaultDBName = uuid();
+    getConnectedClient = () =>
+      Promise.resolve(getDummyClient({ defaultDBName }));
```
oskardudycz, line 27: **COMMENT:** I'm using a dummy client, not to mock it, but the storage resolution doesn't require the real MongoDB; it's only based on passing proper parameters to MongoDB API. I can then check if the collection was indeed created, but I'm not trying to mock everything, just stubbing the dummy setup.

src/packages/emmett-mongodb/src/testing/mongoMocks.ts
```diff
@@ -0,0 +1,42 @@
+import type {
+  Collection,
+  CollectionOptions,
+  Db,
+  DbOptions,
+  Document,
+  MongoClient,
+} from 'mongodb';
+
+export const getDummyCollection = <TSchema extends Document = Document>(
```
oskardudycz, line 10: **COMMENT:** Intention behind this dummy collection is to be used in the structure tests, not to act as an in-memory collection.

alex-laycalvert (approved):
I like it!

alex-laycalvert:
I think this should be enough for a while as far as how far the store options can be customized.

oskardudycz:
@alex-laycalvert, thanks for checking it out!

> I think this should be enough for a while as far as how far the store options can be customized.

Yes, I think that we now have good foundations of the MongoDB storage layout 👍 

I'll release it next week, and also write an article about it 🙂

PR #152, opened 2024-12-18
Added onAfterCommit hook for in-process notifications after events were appended

oskardudycz:
Added `DefaultEventStoreOptions` with `onAfterCommit` hook for in-process notifications after events were appended. All stores will eventually have to apply them.

 The `onAfterCommit` hook will be called **AFTER** events were stored in the event store.
It's designed to handle scenarios where delivery and ordering guarantees do not matter much.

## WARNINGS

1. It will be called **EXACTLY ONCE** if append succeded.
2. If the hook fails, its append **will still silently succeed**, and no error will be thrown.
3. Wen process crashes after events were committed, but before the hook was called, delivery won't be retried. That can lead to state inconsistencies.
4. In the case of high concurrent traffic, race conditions may cause ordering issues. For instance, where the second hook takes longer to process than the first one, ordering won't be guaranteed.

This PR adds implementation to InMemoryEventStore, and provides basic capabilities like `tryPublishMessagesAfterCommit` helper and `forwardToMessageBus` method to make easier integration with in-process message bus.

## Example Usage

You can now use:

```ts
const eventStore = getInMemoryEventStore({
  hooks: {
    onAfterCommit: (events) => {
      // bring your own handling
      console.log(JSON.stringify(events));
    },
  },
});
```

Or integrate with [InMemoryMessageBus](https://event-driven.io/en/inmemory_message_bus_in_typescript/).

```ts
const messageBus = getInMemoryMessageBus();
messageBus.subscribe((event: TestEvent) => {
  // bring your own handling
  console.log(JSON.stringify(event));
}, 'test');

const eventStore = getInMemoryEventStore({
  hooks: {
    onAfterCommit: forwardToMessageBus(messageBus),
  },
});
```

### DIFF
--- a/src/packages/emmett/src/eventStore/afterCommit/afterEventStoreCommitHandler.ts
+++ b/src/packages/emmett/src/eventStore/afterCommit/afterEventStoreCommitHandler.ts
@@ -1,9 +1,5 @@
 import { type Event, type ReadEvent } from '../../typing';
-import type {
-  DefaultEventStoreOptions,
-  EventStore,
-  EventStoreReadEventMetadata,
-} from '../eventStore';
+import type { EventStore, EventStoreReadEventMetadata } from '../eventStore';
 
 type AfterEventStoreCommitHandlerWithoutContext<Store extends EventStore> = (
   messages: ReadEvent<Event, EventStoreReadEventMetadata<Store>>[],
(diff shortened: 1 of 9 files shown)


oskardudycz:
@alex-laycalvert FYI, we were discussing it how to publish events from MongoDB EventStore in process. This is the first step to achieve that. It'll also need to be added to the MongoDB Event Store. I'll also similarly add more advanced handling to PostgreSQL Consumer.

PR #154, opened 2024-12-20
Added emmet-sqlite to workspace

davecosec:


oskardudycz (changes requested):
@davecosec, could you also add sqlite package to:
- tsconfig - https://github.com/event-driven-io/emmett/blob/ef8e32b771bcf78785d9aa6bd574d6e221104c27/src/tsconfig.json#L32
- eslint config: https://github.com/event-driven-io/emmett/blob/ef8e32b771bcf78785d9aa6bd574d6e221104c27/src/eslint.config.mjs#L96

to complete setup?

The package-lock changes indeed looks suspicious, but might be that those are needed sqlite3 dependencies or peer dependencies 🤷

davecosec:
> The package-lock changes indeed looks suspicious, but might be that those are needed sqlite3 dependencies or peer dependencies 🤷

https://github.com/TryGhost/node-sqlite3/issues/1800 

hrm..

oskardudycz:
> > The package-lock changes indeed looks suspicious, but might be that those are needed sqlite3 dependencies or peer dependencies 🤷
> 
> [TryGhost/node-sqlite3#1800](https://github.com/TryGhost/node-sqlite3/issues/1800)
> 
> hrm..

Let’s keep it as it is, for now, and later consider switching to better sqlite3 if it appears better maintained and fulfilling our needs.

oskardudycz (approved):

oskardudycz:
Thanks @davecosec, I’m pulling that in 👌

PR #159, opened 2024-12-29
Updated API tests to use strongly-typed Event Store type

oskardudycz:
Previously, it was always taking the Event Store implementation, which was forcing to do casting while setting up, which was not great.

Fixed also Event Store wrapper to just override the needed methods passing the rest as they are. Without it, even store extensions (like MongoDB projections API) wouldn't work.

Removed event store wrapper usage from E2E tests, as it was redundant: checks are only made on responses.


oskardudycz:
@thiagomini FYI.

PR #161, opened 2024-12-29
Added the first version of Pongo inline projection specification

oskardudycz:
There are no changes in the projection logic; I just moved the code from the index.ts to mongoDBInlineProjection.ts.


src/packages/emmett-mongodb/src/eventStore/projections/mongoDBInlineProjection.ts
```diff
@@ -0,0 +1,210 @@
+import {
```
oskardudycz, line 1: **COMMENT:** Nothing to see here, Git is not showing that it was just moved from index.ts 🤷‍♂️

oskardudycz:
@alex-laycalvert, if you find time, I'd be grateful if you could look into that. If you don't, that's fine, we can fine-tune it later :+1:

alex-laycalvert (approved):

PR #165, opened 2025-01-01
Added SQLite append to stream

davecosec:


src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
```diff
@@ -0,0 +1,213 @@
...
+        eventId: uuid(),
+        streamPosition: BigInt(i + 1),
+        ...e.metadata,
+      },
+    }),
+  );
+
+  let result: AppendEventResult;
+
+  await db.command(`BEGIN TRANSACTION`);
```
davecosec, line 57: i dont like this, no checks in case someone wants to wrap in a transaction

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.int.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+      uuid(),
+      'shopping_cart',
+      events,
+      {},
+    );
+
+    assertTrue(result.success);
+    assertEqual(result.nextStreamPosition, 2n);
+    assertIsNotNull(result.lastGlobalPosition);
+    assertTrue(result.lastGlobalPosition > 0n);
```
oskardudycz, line 85: **SUGGESTION:** I think that we could check here explicitly for 1n, as above we know that the next one is 2n.

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
oskardudycz: We can cover that later :+1:

src/packages/emmett-sqlite/src/eventStore/schema/tables.ts
```diff
@@ -26,9 +26,30 @@ export const eventsTableSQL = sql(
...
       created                DATETIME                  DEFAULT CURRENT_TIMESTAMP,
-      PRIMARY KEY (stream_id, stream_position, partition, is_archived)
+      UNIQUE (stream_id, stream_position, partition, is_archived)
   ); 
 `,
 );
+
+export const eventStreamTrigger = sql(
+  `
+    CREATE TRIGGER emt_global_event_position
```
oskardudycz, line 38: **FIX:** Personally, I'd prefer to have that done explicitly in the append function rather than trigger. Triggers sounds like the automated way, but it'll be much harder to track what's happening.

Especially that we should use upsert to streams table in the future for the optimistic concurrency check.

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
```diff
@@ -0,0 +1,264 @@
...
+      const streamPosition =
+        BigInt(event.metadata.streamPosition) + BigInt(expectedStreamVersion);
+
+      queryBuilder.parameterMarkers.push(`(?,?,?,?,?,?,?,?,?)`);
+      queryBuilder.values.push(
+        streamId,
+        streamPosition.toString() ?? 0,
+        partition ?? defaultTag,
+        JSONParser.stringify(event.data),
+        JSONParser.stringify({ streamType: streamType, ...event.metadata }),
```
oskardudycz, line 232: **QUESTION:** Is there a specific reason why do we always add stream type to metadata?

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
```diff
@@ -0,0 +1,264 @@
...
+  return error?.errno !== undefined && error.errno === 19;
+};
+
+async function getLastStreamPosition(
+  db: SQLiteConnection,
+  streamId: string,
+  expectedStreamVersion: bigint | null,
+): Promise<bigint> {
+  const result = await db.querySingle<{ stream_position: string } | null>(
+    `SELECT CAST(MAX(stream_position) AS VARCHAR) AS stream_position FROM ${streamsTable.name} WHERE stream_id = ?`,
```
oskardudycz, line 189: **CONCERN:** I think that this MAX is redundant here, as we should always have a single record.

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.int.spec.ts
```diff
@@ -0,0 +1,265 @@
+import {
```
oskardudycz, line 1: **KUDOS:** Nice set of tests :+1:

oskardudycz (changes requested):
@davecosec thank you for the change. If trigger goes away, I'm happy to get this in :) See also my other comments.

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
davecosec: oh it was to get it into the streams table via the trigger but that's gone now

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
davecosec: yeah totally redundant, can't remember why that happened

src/packages/emmett-sqlite/src/eventStore/schema/tables.ts
```diff
@@ -26,9 +26,9 @@ export const eventsTableSQL = sql(
       event_type             TEXT                      NOT NULL,
       event_id               TEXT                      NOT NULL,
       is_archived            BOOLEAN                   NOT NULL DEFAULT FALSE,
-      global_position        BIGINT                    ,
+      global_position        INTEGER                   PRIMARY KEY,
```
oskardudycz, line 29: **FIX:** This should remain bigint, as we may have a lot of events.

oskardudycz: NVM, I see that SQLite doesn't have BIGINT, and that's only alias to Integer https://www.sqlite.org/datatype3.html

oskardudycz: Interesting:

> If a table has the primary key that consists of one column, and that column is defined as INTEGER then this primary key column becomes an alias for the rowid column.
>
> Notice that if you assign another integer type such as BIGINT and UNSIGNED INT to the primary key column, this column will not be an alias for the rowid column.

[Source](https://www.sqlite.org/datatype3.html).

SQLite seems to be _interesting_ database.

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
```diff
@@ -0,0 +1,277 @@
...
+      global_position: string;
+    } | null>(sqlString, values);
+
+    if (returningId?.global_position == null) {
+      throw new Error('Could not find global position');
+    }
+
+    globalPosition = BigInt(returningId.global_position);
+
+    const positions = await db.querySingle<{ stream_position: string } | null>(
```
oskardudycz, line 136: **FIX:** We need to check whether the version inside the streams table is still the same as expected. We can also do the exact operation based on the expected version value. If it's `0`, then that means we should insert it. If it's bigger, then we should do an update. That guards from race conditions where someone managed to append new events between we read the last position and update it. If we were just incrementing it as here, we could not detect it.

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    const { sqlString, values } = buildEventInsertQuery(
+      events,
+      expectedStreamVersion,
+      streamId,
+      streamType,
+      options?.partition?.toString() ?? defaultTag,
+    );
+
+    const returningId = await db.querySingle<{
```
oskardudycz, line 126: **FIX:** The attempt to insert events should be after the update to stream, as it's more consuming. If updating the stream position fails on version mismatch, then we can fail fast and not even try to append new events.

oskardudycz (changes requested):
@davecosec we're getting there, thanks for doing the fixes. I added two more comments. I'd be grateful if you could have a look on them 🙂

src/packages/emmett-sqlite/src/eventStore/schema/appendToStream.ts
oskardudycz: @davecosec my guess is that was a leftover from calculating global positon using MAX :slightly_smiling_face:

oskardudycz (approved):
@davecosec, looks great now! Thanks for patiently adjusting to my suggestions :+1:

Issue #169, opened 2025-01-09
Add the first version of MongoDB async subscriptions and projections with at-most-once delivery


oskardudycz:
@alex-laycalvert  Here's the recommended break-down, but I'm open to discussion:

## 1. Add the first onAfterCommit hook to MongoDBEventStore

This could be done similarly to what I did for the in-memory one: https://github.com/event-driven-io/emmett/pull/152. 
 
That should also enable using the in-memory bus.

## 2. Define the `MongoDBEventStoreConsumer`  and `MongoDBSubscription`

You can check PostgreSQL consumer for inspiration (see [here](https://github.com/event-driven-io/emmett/blob/main/src/packages/emmett-postgresql/src/eventStore/subscriptions/postgreSQLEventStoreConsumer.int.ts)). Compared to it, the options should take MongoDBEventStore for now. Thanks to that, the consumer would plug into the onAfterCommit hook. Of course, you don't need to implement polling, as it'll allow you to listen only to upcoming events.

So it'll be just a wrapper, but it'll allow us to plug async projections based on Oplog or change streams later on. 

## 3. Define async projections

Async projections could be plugged as syntactic sugar on top of subscriptions. Subscriptions can have in context mongoClient, which will be useful for passing them to projections.

We'll probably have a similar number of projections as for PostgreSQL inline projections, with the most generic one taking just event(s) and context.

Then, multistream and single stream as Pongo projections 

You can check:
- [PostgreSQL projection](https://github.com/event-driven-io/emmett/blob/main/src/packages/emmett-postgresql/src/eventStore/projections/index.ts)
- [Pongo projection built on top of it](https://github.com/event-driven-io/emmett/blob/main/src/packages/emmett-postgresql/src/eventStore/projections/pongo/projections.ts)
- [Article on using Pongo projections](https://event-driven.io/en/emmett_projections_testing/)

You can also check Pongo handle method  that detect changes made to documents (see [here](https://github.com/event-driven-io/Pongo/blob/1a7e6f56454de6f25fa1e58aacf66b3044a434d0/src/packages/pongo/src/core/collection/pongoCollection.ts#L338 )) use Pongo to handle operations (whether insert, update or delete). It could be emulated to MongoDB in almost the same way: 

## Notes

I suggest doing it like that, as I have the general idea of async handling to have the following concepts:
- **Consumer** - responsible for getting the messages from source (e.g. MongoDB, PostgreSQL event store, later on Kafka, SQS, etc.)
[...]

PR #170, opened 2025-01-09
Partial #169 - MongoDB EventStore `onAfterCommit` hook

alex-laycalvert:
## Description

- Adds a basic `onAfterCommit` hook to the `MongoDBEventStore` options modeled after the in-memory event store
- Adds same test suite for the `inMemoryEventStore.onAfterCommit` test
- ~~Adjusts the typing of `EventStoreReadEventMetadata` to allow for an undefined `GlobalPosition`. Without this change, attempting to pass the events to `tryPublishMessagesAfterCommit` gave a TS error because the events didn't have a global position, even though we didn't need one for mongodb~~

## TODO

- [x] Tests

Relates to #169


alex-laycalvert:
@oskardudycz the change to the metadata typing is now giving an error for the in memeory event store, I'll check it out and if it gets too complicated I can just typecast the usage in mongodb

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
```diff
@@ -388,6 +393,15 @@ class MongoDBEventStoreImplementation implements MongoDBEventStore, Closeable {
       );
     }
 
+    await tryPublishMessagesAfterCommit<MongoDBEventStore>(
+      // @ts-expect-error Issues with `globalPosition` not being present causing the type for metadata to expect `never`
```
oskardudycz, line 397: @alex-laycalvert, that's weird; I noticed that TS got crazy on my other computer, but when I reinstalled fresh packages, then it worked fine. I'm not sure if that's the same issue, but I'll check the typing after the merge :+1:

oskardudycz (approved):
@alex-laycalvert looks good to me, thank you!

src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.ts
oskardudycz: Yes, because technically, it's not `undefined` but optional. TypeScript is treating those cases differently. I'll have a look, it might be that I messed something out.

PR #171, opened 2025-01-09
Added read stream

davecosec:
Added ability to read stream


oskardudycz (approved):
Sweet, I'm pulling that in! Now, I'm keeping my fingers crossed for applying it to the event store!

PR #175, opened 2025-01-16
Fixed browser compatibility test

oskardudycz:
For some reason, `uuid` peer dependency started to fail for skypack CDN. 

Switched from it to jsdelivr.net and locked to the specific version.


oskardudycz:
@devcosec fyi

PR #177, opened 2025-01-21
Fixed typing for Command Handler async result

oskardudycz:
Without that, it wouldn't be possible to wrap decision logic with async middleware.


oskardudycz:
@alex-laycalvert FYI, we were discussing that during the workshop.

PR #184, opened 2025-02-08
Renamed events to messages in the PostgreSQL and SQLite schemas and added message kind column

oskardudycz:


oskardudycz:
@davecosec FYI.

oskardudycz:
Okay, I'm pulling that in. I checked if the PostgreSQL migration would work on the sample project, and it did.

Issue #179, opened 2025-01-22
Add inline projections to In Memory event store


oskardudycz:
Here's the recommended outline:
## 1. Add inline projection setup to InMemoryEventStore options

That can be done in a similar way as in the [PostgreSQL event store options](https://github.com/event-driven-io/emmett/blob/52a3ba72359c4ae2de1a041c7dac3d74f77f0a63/src/packages/emmett-postgresql/src/eventStore/postgreSQLEventStore.ts#L131).

Something like:

```ts
export type InMemoryReadEventMetadata = ReadEventMetadataWithGlobalPosition;

export type InMemoryProjectionHandlerContext = {
  eventStore: InMemoryEventStore
};

export type InMemoryEventStoreOptions = DefaultEventStoreOptions<InMemoryEventStore> & {
  projections?: ProjectionRegistration<
    'inline',
    InMemoryReadEventMetadata,
    InMemoryProjectionHandlerContext
  >[];
};
````

Then, [cache them in a similar fashion as Postgres projections](https://github.com/event-driven-io/emmett/blob/52a3ba72359c4ae2de1a041c7dac3d74f77f0a63/src/packages/emmett-postgresql/src/eventStore/postgreSQLEventStore.ts#L171).

- [x] **DONE IN:** #186 

## 2. Add InMemoryDatabase

You can clone this one for now: https://github.com/oskardudycz/EventSourcing.NodeJS/blob/main/workshops/introduction_to_event_sourcing/src/11_projections_single_stream/tools/database.ts.

Replace `JSON.parse` and `JSON.stringify` with [Emmett JSONParser](https://github.com/event-driven-io/emmett/blob/52a3ba72359c4ae2de1a041c7dac3d74f77f0a63/src/packages/emmett/src/serialization/json/JSONParser.ts)

- [x] **DONE IN:** #186 

## 3. Port `handle` method to InMemoryDatabase based on Pongo one

It's placed here, should be copy/paste with small adjustments: https://github.com/event-driven-io/Pongo/blob/1a7e6f56454de6f25fa1e58aacf66b3044a434d0/src/packages/pongo/src/core/collection/pongoCollection.ts#L338.

- [x] **DONE IN:** #207, #226   

## 4. Add similar projections like in Pongo

See them: https://github.com/event-driven-io/emmett/blob/52a3ba72359c4ae2de1a041c7dac3d74f77f0a63/src/packages/emmett-postgresql/src/eventStore/projections/pongo/projections.ts

[...]

PR #186, opened 2025-02-10
In memory inline projections - Add inMemoryDatabase

stepaniukm:
This PR covers the first two steps of this issue: https://github.com/event-driven-io/emmett/issues/179

### DIFF
--- a/src/packages/emmett/src/database/inMemoryDatabase.ts
+++ b/src/packages/emmett/src/database/inMemoryDatabase.ts
@@ -1,0 +1,49 @@
+import { JSONParser } from '../serialization';
+
+export interface DocumentsCollection<T> {
+  store: (id: string, obj: T) => void;
+  delete: (id: string) => void;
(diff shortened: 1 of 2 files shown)


src/packages/emmett/src/database/inMemoryDatabase.ts
```diff
@@ -0,0 +1,49 @@
+import { JSONParser } from '../serialization';
+
+export interface DocumentsCollection<T> {
```
oskardudycz, line 3: **QUESTION:** @stepaniukm, is it intentional that members in this file are not exposed in the main index file? I'm fine with not exposing them yet until it's ready; I'm just double-checking.

stepaniukm: Yup that is intentional, first I want to complete migration to something actually usable, tested and more stable, before exposing it publicly. One little thing I removed type `DocumentHandler` that will be a part of future effort.

PR #187, opened 2025-02-11
Added optional getDocumentId selector for single stream projection

oskardudycz:
Thanks to that, one can still use a custom ID based on the event data.

I'll need to think how to make it possible to assign the `_id` in the evolve function, but that will require changing the Pongo handle method. So it'll come in the follow-up.

@jameswoodley FYI


jameswoodley (approved):

jameswoodley:
Thanks!

PR #193, opened 2025-02-14
Added default event type to consumers

oskardudycz:
Thanks to that, processors will be automatically typed. This can be useful if we have multiple processors for the same stream subscription.

It's optional and falls back to the event, so each processor can still have its own event type (e.g. when subscribing to all events).

Another option is to put a union of all events in consumer and then use just a subset of those types in processors.


oskardudycz:
@mbirkegaard fyi

PR #199, opened 2025-02-24
Add API reference link to documentation sidebar and update title in API docs

tburny:
The API docs page is actually a kind of reference documenation that provides a deep dive into the intrinsics of Emmet. Change the page title to reflect this change and make the page accessible via the sidebar navigation.

Note: I did not rename the file, as cool URIs dont change. However I am planning to refactor the page into several parts for improved developer experience in the future, so it might simply redirect to a new page at some point.

### DIFF
--- a/src/docs/.vitepress/config.mts
+++ b/src/docs/.vitepress/config.mts
@@ -27,7 +27,10 @@ export default defineConfig({
     sidebar: [
       {
         text: 'Documentation',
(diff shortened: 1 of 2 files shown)


oskardudycz:
The reason why I removed the API Reference was that, for now I didn't have time to provide the proper setup and description, but as you want  to make some changes there, I'm happy to take it in :+1:  @tburny thanks for doing that :+1:

PR #200, opened 2025-02-24
Add overview page to documentation

tburny:
The overview page provided users with a map to navigate the documentation while it keeps growing with the project.

While currently there are only two pages (Getting Started and API docs), this is going to change in the near future with further refactorings and improvements to the documentation.

This PR also introduces Diataxis - without imposing a folder structure at the moment to allow for organic growth.

### DIFF
--- a/src/docs/.vitepress/config.mts
+++ b/src/docs/.vitepress/config.mts
@@ -17,7 +17,7 @@ export default defineConfig({
         items: [
           { text: 'Overview', link: '/overview' },
           { text: 'Getting started', link: '/getting-started' },
-          { text: 'API reference', link: '/api-reference' },
+          { text: 'API reference', link: '/api-docs' },
         ],
       },
       { text: '🧑‍💻 Join Discord Server', link: 'https://discord.gg/fTpqUTMmVa' },


tburny:
Just FYI: I created the overview page by looking at the [Django documentation](https://docs.djangoproject.com/en/5.1/), which also uses Diataxis. Always good to have a reference.

src/docs/overview.md
```diff
@@ -0,0 +1,36 @@
+# Overview
+
+Everything you need to know about Emmet and Event Sourcing.
```
oskardudycz, line 3: **COMMENT:** Later on in Overview, we should probably do more information like I did once in Marten (so what it is and main features) https://martendb.io/introduction.html#what-is-marten.

But this can go as a follow-up.

src/docs/overview.md
```diff
@@ -0,0 +1,36 @@
...
+## Getting help
+
+If you need help or got stuck, feel free to ask on the [Emmet Community Discord Server](https://discord.gg/fTpqUTMmVa).
+
+## API reference
+
+The [API reference](/api-docs) provides you with definitions and insights of Emmets core building blocks:
+
+- **Events** are the centerpiece of event-sourced systems. They represent both critical points of the business process but are also used as the state.
+- **Commands** represent the intent to record a business operation.
```
oskardudycz, line 20:
```suggestion
- **Commands** represent the intent to perform a business operation.
```

src/docs/overview.md
```diff
@@ -0,0 +1,36 @@
...
+- **Event Store** for recording events
+- **Command Handlers** aggregate several events from the event store into a single object to perform the intended business operation, resulting in one or more events to record the change.
```
oskardudycz, line 22:
```suggestion
- **Command Handlers** are responsible for handling business logic. They read events from the event store for specific events and build a decision model. Then, check business rules and perform the intended business operation, resulting in one or more events to record the change.
```

src/docs/overview.md
```diff
@@ -0,0 +1,36 @@
...
+
+## How the documenantation is organized
+
+Currently documentation for Emmet is spread across several places: This website, the Emmet Discord and quite a few blog articles.
```
oskardudycz, line 26: **COMMENT:** Maybe we could link those articles here? Thoughts. And tell something along the lines:

> You can also find more information about Emmett in those articles. We're working on including all the knowledge centralised in this website.

Then, what you did below.

@tburny, thoughts?

oskardudycz (approved):
@tburny thank you for doing that, I added small comments and suggestions. After they're resolved, I'm more than happy to pull that in.

src/docs/overview.md
tburny: I'd rather keep the overview page concise. My idea would be to have "Further reading sections" at the bottom of the documentation pages that link to the respective blog articles relevant in that context.

As a first step we can have such a section in the overview and migrate them from there.

tburny: Just added the link in a "Further reading" section so they are at least somewhere in the documentation.

tburny:
I implemented the changes as requested (i.e. spelling Emmett correctly). Also I added some minor improvements:
- adding the blog links as "further reading" section and linking to the section from within the "Overview" page
- add links to Discord in the overview page
- There was no way from the landing page to the overview page, so I simply added it to the main nav and made "getting stated" a menu item there. This way new users will probably click the "getting stated" link (call to action), while if I am looking for something else I can still use the menu to roughly navigate to my area of interest (e.g. API reference).

Ready for the second round!

src/docs/overview.md
oskardudycz: Makes sense to me, thanks for doing it :+1:

oskardudycz:
@tburny I think that for now, it makes sense to route people directly to getting started, later on, we can enhance it when we have more content 🙂

Thanks again for sending those docs updates; they are much appreciated!

tburny:
Looks like we were a bit too eager to merge this. Somehow the merge checks failed and vitepress is reporting dead links: https://github.com/event-driven-io/emmett/actions/runs/13618738966/job/38065200339

Issue #196, opened 2025-02-18
Refactor evolve and evolveWithMetadata to avoid code duplication and a potential bug for amounts


shraddha38:
### Summary:
The evolve and evolveWithMetadata functions in [ShoppingCart](https://github.com/event-driven-io/emmett/blob/072c09b0d44789fd365acfaa216ade79cfce53cf/src/packages/emmett-tests/src/eventStore/shoppingCart.domain.ts#L39) have some code duplication and lack safety for new event types.

## Suggested Improvements

1. Removing duplication - Reduce duplication in event handling logic and ensure safety by adding a default case in switch statements

Same logic is used in evolve and evolveMetatadata fn, so we can extract it as following - 

```ts
const updateCart = (state: ShoppingCart, productItem: PricedProductItem): ShoppingCart => ({
  productItems: [...state.productItems, productItem],
  totalAmount: state.totalAmount + productItem.price * productItem.quantity,
});
```

2. Bug- Prevent discounts from causing negative amounts

> Current 
```ts
 case 'DiscountApplied':
      return {
        ...state,
        totalAmount: state.totalAmount * (1 - data.percent / 100),
      };
```
> Suggestion

```ts
 case 'DiscountApplied':
      return {
        ...state,
        totalAmount: Math.max(state.totalAmount * (1 - data.percent / 100), 0),
      };
```

## Next Steps

1. Need review from maintainers on this issue. 
2. After that I can create a Pr for the changes.

shraddha38:
@oskardudycz Thanks, totally aligned to keep the code as is. Added a small pr for the max to avoid a negative amount for discounts. https://github.com/event-driven-io/emmett/pull/201

PR #201, opened 2025-02-26
chore: add max for discounts

shraddha38:
https://github.com/event-driven-io/emmett/issues/196

## Summary 
Bug- Prevent discounts from causing negative amounts

Current Behaviour-

```ts 
 case 'DiscountApplied':
      return {
        ...state,
        totalAmount: state.totalAmount * (1 - data.percent / 100),
      };
 ```

### DIFF
--- a/src/packages/emmett-tests/src/eventStore/shoppingCart.domain.ts
+++ b/src/packages/emmett-tests/src/eventStore/shoppingCart.domain.ts
@@ -38,7 +38,7 @@ export const evolve = (
     case 'DiscountApplied':
       return {
         ...state,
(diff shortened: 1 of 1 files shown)


oskardudycz:
@shraddha38 looks good to me, but could you run `npm run fix` in your code? This will apply linting, if you're using vscode it should be in theory automatically enabled on save. See also https://github.com/event-driven-io/emmett/blob/main/CONTRIBUTING.md#code-style

oskardudycz:
Thanks @shraddha38, feel also invited to join our Discord channel: https://discord.gg/fTpqUTMmVa

That's the place where it's probably the fastest to discuss ideas 🙂

PR #203, opened 2025-03-01
Move entries from api-guide page into their own distinct pages in /api-reference

tburny:
With a growing number of types and definitions, the API reference page becomes quite lengthy and it is easy getting lost while reading.

Changes:
- The api-guide.md contents were moved into separate pages for Event, Command, EventStore and CommandHandler.
- Each API reference item now is its own item in the sidebar (easy to find)
- Add overview page at /api-reference. Clicking the "Next" link links to /api-reference/event as first page from the navigation
- For each page, use a similar structure of: _Description text ("Overview")_, _Usage, Definition, See also. This "template" structure makes it easier to find the thing one is looking for at first glance. The sidebar nav on the right supports this (looks quite the "same" for Event, Command, Event Handler and Command Handler)
- Keep the api-docs.md page and link to the moved contents, as [Cool URIs dont change](https://www.w3.org/Provider/Style/URI).

Open items (tbd):
- [x] Naming: Is it command-handler.md or `commandhandler.md`? I am favouring the second as reference to the type name in the code
- [x] Decide on order of Usage and Definition. While I think it is more convenient to skip the code definitions and see "how can I use it" first, it feels a bit strange like in "First definition, then usage".


src/docs/api-docs.md
```diff
@@ -6,98 +6,14 @@ outline: deep
...
-The type is a simple wrapper to ensure the structure's correctness. It defines:
-
-- **type** - event type name,
-- **data** - represents the business data the event contains. It has to be a record structure; primitives are not allowed,
-- **metadata** - represents the generic data event contains. It can represent telemetry, user id, tenant id, timestamps and other information that can be useful for running infrastructure. It has to be a record structure; primitives are not allowed.
-
-See more context in [getting started guide](./getting-started.md#events)
-
-<<< @./../packages/emmett/src/typing/event.ts
+[Moved here](/api-reference/event)
```
oskardudycz, line 9: **QUESTION:** Do we need to keep this file?

oskardudycz:
Regarding open items: 

As mentioned on Discord, I think I'd prefer snake or kebab case, as those files will be mapped to URLs and that will make them easier to find, and potentially improve SEO. Yet I'm not strong on that, as we can always later on change it and add url rewrites.

Regarding order of Usage and Definition. I like to teach in a way that I give the problem statement, then basic solution, and then, so:
1. Problem statement,
2. Potential solution so usage,
3. Explaining why this solution is recommended and definition.

But I'm also not strong on that. That works for me, as it helps to keep readers motivated to read. A bit of storytelling technique is always good. If we have a full definition first, then the solution, many people immediately scroll down.

Still, we can also follow Diataxis' advice and use different patterns depending on the intention.

In general, what you did seems legit to me.

oskardudycz:
Ah, @tburny could you run `npm run fix` to clean up some linter issues with md files?

tburny:
I fixed the issues and kept things as-is, for the aforementioned reasons.

oskardudycz (approved):

oskardudycz:
Thanks @tburny !

PR #206, opened 2025-03-02
WIP: In memory database handle

stepaniukm:
This PR introduces the first iteration of in memory database that is more compatible with Pongo, than the first tries. It is supposed to be the 3rd step from the issue: https://github.com/event-driven-io/emmett/issues/179


stepaniukm:
This PR was broken because I made a git mistake, correct one is to be found in the next one -> #207

PR #209, opened 2025-03-03
feat: add devcontainer configuration for consistent development environment

tburny:
This pull request adds a `.devcontainer` configuration, enabling developers to easily set up a consistent development environment using Docker containers. The configuration ships with the necessary VSCode plugins like Prettier and ESlint - all batteries included.

Key benefits:

- **Simplified Setup:** Provides a pre-configured environment, reducing setup time and eliminating dependency conflicts.
- **Uniform Development:** Ensures a consistent environment across local machines and GitHub Codespaces.
- **Improved Onboarding:** Facilitates quicker onboarding for new contributors by providing a ready-to-use development environment.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,1 @@
+FROM mcr.microsoft.com/devcontainers/javascript-node:1-22-bookworm
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,14 +3,7 @@
 {
   "image": "mcr.microsoft.com/devcontainers/javascript-node:1-22-bookworm",
   "name": "Node.js",
-  // Fixes permissions problems when using podman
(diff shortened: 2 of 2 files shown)


tburny:
There is still an incompatiblity between docker and podman that I need to investigate. The focus should be that the configuration works for GitHub Codespaces.

oskardudycz:
Wow, that's cool stuff. I have always wanted to add it! It would be also great if we added some note in README and CONTRIBUTING guide, but that can come later :)

oskardudycz:
> Podman requires a command-line argument --userns=keep-id with a mapping, but the value is incompatible with Docker (which only supports host).Any other means of trying to pass the value in a podman-specific way were unsuccessful. For podman, the value "host" sets uid=0 and gid=0, effectively meaning "root". This way the workspace folder was created / mounted as "root" and the image runs as "node" user, who cannnot write the workspace folder as it is owned by root:root.

Might be that we can document-around-it for folks using Podman. Probably that's still less people than those using Docker, so at least locally, they could make that work. Or we could provide different setup for them if it's possible to switch, or run different shell.

PR #210, opened 2025-03-03
fix: Eliminate dead links in vitepress

tburny:
This eliminates the dead links amended by the docs:build step, so the build passes.


oskardudycz:
@tburny Thanks for fixing that. I saw the build failure and wanted to handle it in the morning, but you were faster. Later today, I will make docs build a part of the regular PR pipeline. That should help to detect it before merging.

oskardudycz (approved):

PR #211, opened 2025-03-03
Added eventStore onBeforeCommit hook test

davecosec:


src/packages/emmett-sqlite/src/eventStore/SQLiteEventStore.e2e.spec.ts
```diff
@@ -248,6 +248,35 @@ void describe('SQLiteEventStore', () => {
     assertIsNotNull(stream.events);
     assertEqual(1, stream.events.length);
   });
+
+  void it('should allow events to be processed in the onBeforeCommit hook', async () => {
```
oskardudycz, line 252: **SUGGESTION:** @davecosec, could you also add test checking if append fails, when `onBeforeCommit` code throws exception?

PR #214, opened 2025-03-06
[`MongoDBEventStore`] Feature - Basic implementation of `MongoDBEventStoreConsumer`

alex-laycalvert:
## Description

- My attempt at implementing `MongoDBEventStoreConsumer`. The current impl. is just a basic pub/sub pattern and you can call `consumer.publish` inside of the `onAfterCommit` hook for your event store which will allow you to actually utilize the consumer.
- Adds `eventStore` to the `onAfterCommit` handler context for convenience

I know the postgres impl. of consumers is more involved with batching, so there's definitely room to make this more robust.

### DIFF
--- a/src/packages/emmett-mongodb/src/eventStore/consumers/index.ts
+++ b/src/packages/emmett-mongodb/src/eventStore/consumers/index.ts
@@ -1,0 +1,1 @@
+export * from './mongoDBEventStoreConsumer';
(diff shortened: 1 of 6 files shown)


oskardudycz (changes requested):
@alex-laycalvert, as a general concept that looks good to me, thank you for doing that.

Yet, could you align the API with the current Consumers =>Processors API? See the EventStoreDB consumer as an example: https://github.com/event-driven-io/emmett/blob/4f56d0794a02ba24a761c2ef9ee020b7e2a64cba/src/packages/emmett-esdb/src/eventStore/consumers/eventStoreDBEventStoreConsumer.handling.int.spec.ts#L69-L81

It should be probably closer than PostgreSQL implementation.

I'm fine with starting with non-batched option, but I'd like to have at leas the API look similar, as that will need to be aligned eventually (and sooner than later, as I'm working on the common typing abstractions on that ATM in PR: https://github.com/event-driven-io/emmett/pull/202

If the consumers and processors concept is unclear, I'm happy to explain that deeper here on on Discord.

PR #220, opened 2025-03-18
feat: Add quick start guide

tburny:
Add a quick start guide as a tutorial. It guides the user through the fundamental steps to build a basic Emmett application.

As a learning-oriented tutorial, the quick start guide plays the "tutor" for the readers in order to help them acquire knowledge about Emmett and guiding them through the necessary steps without much thinking.

See https://diataxis.fr/tutorials/ for reference.

### DIFF
--- a/src/docs/quick-start.md
+++ b/src/docs/quick-start.md
@@ -1,0 +1,684 @@
+---
+title: Quick start
+documentationType: tutorial
+---
+
+In this tutorial we will create an application that can increase and decrease a counter. Along this way you will experience the fundamental mechanics and basic building blocks of Emmett.
+
+## Create a Node.js project
+
+Before we can install and use Emmett, we need to set up a Node.js project and add Typescript support.
+To do so, first create a project directory and switch to it:
+
+```sh
+mkdir emmett-quick-start
+cd emmett-quick-start
+```
+
+::: code-group
+
+```sh [npm]
+npm init
+```
+
+```sh [pnpm]
+pnpm init
+```
+
+```sh [yarn]
+yarn init
+```
+
+```sh [bun]
+bun init
+```
+
+:::
+
+For now, just accept the defaults.
+
+> [!NOTE]
+> For the sake of brevity, we will focus on using `npm` as package manager in this tutorial.
+
+The output should look like this:
+
+```text
+package name: (emmet-quick-start)
+version: (1.0.0)
+description:
+entry point: (index.js) dist/index.js
+test command:
+git repository:
+keywords:
+author:
+license: (ISC)
+About to write to /home/tobias/projekte/emmet-quick-start/package.json:
+
+{
+  "name": "emmet-quick-start",
+  "version": "1.0.0",
+  "description": "",
+  "main": "dist/index.js",
+  "scripts": {
+    "test": ""
+  },
+  "author": "",
+  "license": "ISC"
+}
+
+
+Is this OK? (yes)
+```
+
+## Add TypeScript
+
+Now install TypeScript:
+
+```sh [npm]
+npm install typescript @types/node --save-dev
+```
+
+The output should look like this:
+
+```text
+added 3 packages, and audited 4 packages in 1s
+
+found 0 vulnerabilities
+```
+
+Now we initialize TypeScript using
+
+```sh
+npx tsc --init --target ESNEXT --outDir dist --module nodenext --moduleResolution nodenext
+```
+
+The output should look like this:
+
+```text
+Created a new tsconfig.json with:
+
+  target: es2016
+  module: commonjs
+  strict: true
+  esModuleInterop: true
+  skipLibCheck: true
+  forceConsistentCasingInFileNames: true
+
+
+You can learn more at https://aka.ms/tsconfig
+```
+
+Then add the `build:ts` and `build:ts:watch` targets to your package.json:
+
+```json{7-8}
+{
+  "name": "emmet-quick-start",
+  "version": "1.0.0",
+  "description": "",
+  "main": "dist/index.js",
+  "scripts": {
+    "build:ts": "tsc",
+    "build:ts:watch": "tsc -b --watch"
+  },
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@types/node": "^22.13.10",
+    "typescript": "^5.8.2"
+  },
+  "dependencies": {
+    "@event-driven-io/emmett": "^0.34.0"
+  }
+}
+```
+
+Let us add an empty `index.ts` file as entrypoint for our application:
+
+```sh
+touch index.ts
+```
+
+> [!NOTE]
+> In real-life applications, we might want to consider improving the project structure later on. However premature optimization is the root of all evil, so let us start simple and evolve later.
+
+Test your configuration by running
+
+::: code-group
+
+```sh [npm]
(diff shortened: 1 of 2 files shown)


src/docs/quick-start.md
```diff
@@ -0,0 +1,684 @@
...
+::: code-group
+
+```sh [npm]
+npm run build:ts
+```
+
+```text [View output]
+$ npm run build:ts
+
+> emmet-quick-start@1.0.0 build:ts
```
oskardudycz, line 155:
```suggestion
> emmett-quick-start@1.0.0 build:ts
```

src/docs/quick-start.md
```diff
@@ -0,0 +1,684 @@
...
+- Increment the counter
+- Decrement the counter
+
+Let us first create a command to increment the counter in `index.ts` and another to decrement it:
+
+```ts
+import type { Command } from '@event-driven-io/emmett';
+
+// Command<type, payload>
+export type IncrementCounter = Command<
```
oskardudycz, line 227: **FIX:** @tburny, could you use a similar events as in the getting started? I think that if you use simplified ProductItemAdded and ProductItemRemoved, then it'll still be simple enough; you could model a counter of product items, but it'd give already some more real-world examples.

src/docs/quick-start.md
```diff
@@ -0,0 +1,684 @@
...
+  }
+>;
+
+// Union type of all possible counter events (for later)
+export type CounterEvent = CounterIncremented | CounterDecremented;
+```
+
+We use this type to record that our counter has been incremented (or decremented) by the amount in the `by` property.
+
+### Turning commands into events
```
oskardudycz, line 286:
```suggestion
### Recording events
```

src/docs/quick-start.md
tburny: I need to try this out. Actually I adapted the example from the bun test, as it was super short and simple.

However I see the added value when it comes to explaining the advantages of Event Sourcing: https://github.com/AxonIQ/reference-guide/blob/master/architecture-overview/event-sourcing.md

oskardudycz: Yup, it doesn't have to be a full-blown tutorial, as we have in getting started, but something that can give a more relatable hook for ideas, or put a seed :)

tburny: I changed it to a shopping cart :)

src/docs/quick-start.md
tburny: Done

tburny:
@oskardudycz You might want to check this again. I also moved the source code of the tutorial into the snippets folder so we can link to it from the Markdown or anywhere else.

src/docs/quick-start.md
```diff
@@ -0,0 +1,308 @@
...
+Our application needs to perform the following operations:
+
+- Add item: Add an item by its name
+- Remove item: Remove an item by its name
+
+Let us first create a command to add an item in `index.ts` and another to remove one:
+
+<<< @/snippets/quickStart/index.ts#commands
+
+This declares a command whose type is `AddItem` (or `RemiveUten`) and whose payload has a `name` property of type `string`. This `name` property specifies the name of the item that should be added to or removed from the cart.
```
oskardudycz, line 204:
```suggestion
This declares a command whose type is `AddItem` (or `RemoveItem`) and whose payload has a `name` property of type `string`. This `name` property specifies the item's name that should be added to or removed from the cart.
```

src/docs/quick-start.md
```diff
@@ -0,0 +1,308 @@
...
+
+Last but not least, we define a type `ShoppingCartCommand` that accomodates all commands. This way we can easily extend the list of possible commands using union types without having to change the code in other places.
+
+> [!NOTE]
+> The command is only a _request_ to perform the operation, e.g. to add an item to the cart. Depending on the outcome, this request can be successful or not.
```
oskardudycz, line 209:
```suggestion
> The command is only an _intention_ to perform the operation, e.g. to add an item to the cart. Depending on the outcome, this request can be successful or not.
```

src/docs/quick-start.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Recording events
+
+Now we need to put the commands and events together using _command handler_ functions:
+
+<<< @/snippets/quickStart/index.ts#state-definition
+
+This function is responsible for deciding the outcome of the command using business rules (e.g. items may not be added more than once). Currently there are none, so we simply pass the name using the `data.name` property of the `command`.
+
+Emmett promotes the following _decider_ pattern that is easily extensible for more commands:
```
oskardudycz, line 229:
```suggestion
You can group all commands into a unified function that is easily extensible when you add more commands:
```

src/docs/quick-start.md
```diff
@@ -0,0 +1,308 @@
...
+
+<<< @/snippets/quickStart/index.ts#decider
+
+This calls the correct _command handler_ function based on the `type` of the command and returns an event which is going to be _recorded_ to the event store further below.
```
oskardudycz, line 233:
```suggestion
This calls the correct _command handler_ function based on the command's `type`  and returns an event (or an array of events) that will be recorded and stored in the event store further below.
```

src/docs/quick-start.md
```diff
@@ -0,0 +1,308 @@
...
+
+<<< @/snippets/quickStart/index.ts#evolve
+
+Finally we can declare the _command handler_ itself:
+
+<<< @/snippets/quickStart/index.ts#handle
+
+### Putting it all together with an event store
+
+The _Event Store_ records a series of all the events that occurred in our application, similar to a cassette tape.
```
oskardudycz, line 247:
```suggestion
The _Event Store_ is logically a key-value database that records a series of events that happened in our application. It logically groups events in streams. In the traditional approach, a stream represents a record. Each instance of the process or entity will have its own stream. That means that it also has its own ID, and the value is a sequence of events recorded for this process or entity (e.g., all the events for a certain shopping cart).
```

src/docs/quick-start.md
```diff
@@ -0,0 +1,308 @@
...
+
+Starting from the initial state, it collates all events using the `evolve` function (basically left reducing the event stream).
```
oskardudycz, line 249:
```suggestion
Starting from the initial state, it folds all events using the `evolve` function (basically left reducing the event stream).
```

src/docs/quick-start.md
```diff
@@ -0,0 +1,308 @@
...
+---
+ItemAdded(name: Pizza) // ShoppingCart(items: Pizza)
+ItemAdded(name: Ice Cream) // ShoppingCart(items: Pizza, Ice Cream)
+ItemRemoved(name: Ice Cream) // ShoppingCart(items: Pizza)
+```
+
+Why does it look that way?
+Initially there is only one event in the stream, then two, then three, then four. Whenever an event is appended to the stream, _the whole stream is re-read_ to calculate the next state.
+
+One key aspect is that this behavior allows us to change the code for calculating the current state of the shopping cart (e.g. to just count the number of items in the cart) without any negative side effects.
```
oskardudycz, line 288:
```suggestion
One key aspect is that this behavior allows us to change the code for building the current state of the shopping cart (e.g. to just count the number of items in the cart) without any negative side effects.
```

oskardudycz (approved):
Looks great, thanks @tburny !

PR #221, opened 2025-03-18
Fix ESDB consumers

mbirkegaard:
The stream to subscribe from did not get passed, which caused all consumers to subscribe to everything

I'm not sure exactly if this is the right way to solve, but it does fix the issue, where otherwise the stream name passed to `.consumer` would just get ignored.


oskardudycz:
@mbirkegaard thank you for the fix, I had to reopen the other PR in #222, as I couldn't push test fixes to your fork (some permissions issue). It contains your commit.

mbirkegaard:
> @mbirkegaard thank you for the fix, I had to reopen the other PR in #222, as I couldn't push test fixes to your fork (some permissions issue). It contains your commit.

Ah, perfect. Thanks!

PR #207, opened 2025-03-02
add first iteration of inMemoryDatabase with handle possibilities

stepaniukm:
This PR introduces the first iteration of in memory database that is more compatible with Pongo, than the first tries. It is supposed to be the 3rd step from the issue: https://github.com/event-driven-io/emmett/issues/179

PR #226, opened 2025-03-22
InMemoryDatabase: apply fixes on comments from previous PR.

stepaniukm:
This PR is a continuation of the previous one https://github.com/event-driven-io/emmett/pull/207.

Scope of changes:
- Added tests for throwing an error in operationResult when opted-in.
- Added tests for `handle` method on collection.
- Added `find` method on callection to get all results.
- Simplified one if statements 

Regarding `ConcurrencyInMemoryDatabaseError` I decided to create a new Error just because of the porting from https://github.com/event-driven-io/Pongo/blob/main/src/packages/pongo/src/core/typing/operations.ts#L458 that doesn't have matching signature with existing `ConcurrencyError` and I didn't want to drastically reshape that `operationResult`, but I could apply that in next commit if you @oskardudycz think it's better.

Regarding `toFullId` function, I decided for now to not bother figuring out exact behavior of that in inMemoryDatabase implementation with store being simple `Map`. Probably worth revisiting when actually adding different storage possibility.

### DIFF
--- a/src/packages/emmett/src/database/handle.unit.spec.ts
+++ b/src/packages/emmett/src/database/handle.unit.spec.ts
@@ -1,0 +1,202 @@
+import { equal, ok, deepStrictEqual } from 'node:assert/strict';
+import { beforeEach, describe, it } from 'node:test';
+import {
+  type DocumentsCollection,
+  getInMemoryDatabase,
+} from './inMemoryDatabase';
+
+type History = { street: string };
+type Address = {
+  city: string;
+  street?: string;
+  zip?: string;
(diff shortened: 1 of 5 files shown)


oskardudycz (approved):
@stepaniukm looks great, thanks for sending that :+1: 

And yeah, I'd like the Concurrency error be merged, or at least provide some base type for both stream and document concurrency so we have a proper mapping in the express.js middleware to HTTP status.

oskardudycz:
@mbirkegaard fyi.

PR #229, opened 2025-03-27
Added SQLite raw SQL projection tests

davecosec:


src/packages/emmett-sqlite/src/eventStore/projections/index.ts
```diff
@@ -74,14 +73,16 @@ export const sqliteRawBatchSQLProjection = <EventType extends Event>(
     handle: async (events, context) => {
       const sqls: string[] = await handle(events, context);
 
-      for (const sql of sqls) await context.db.command(sql);
+      for (const sql of sqls) await context.connection.command(sql);
     },
   });
 
 export const sqliteRawSQLProjection = <EventType extends Event>(
+  getDocumentId: (event: EventType) => string,
```
oskardudycz, line 81: **CONCERN:** @davecosec, as we discussed, the raw projection should not assume that we need to know the document id. As evolve function is contextual, it could just be creating the SQL based on the event data. `getDocumentId` seems redundant. @davecosec, thoughts, am I missing something?

src/packages/emmett-sqlite/src/eventStore/projections/sqliteRawSQLProjection.single.int.spec.ts
```diff
@@ -0,0 +1,267 @@
...
+
+        const sql = `INSERT INTO 
+          ${projection} 
+          (
+            id, 
+            productItemsCount, 
+            totalAmount, 
+            discountsApplied
+          ) VALUES (
+            "${documentId}",
```
oskardudycz, line 220: **SUGGESTION:** See, as you know, here is precisely the context; then you could just take the document ID based on the streamName or event data. Btw. if you use ReadEvent<EventType>, then metadata will be mandatory, you won't need to use `?` around it.

oskardudycz (changes requested):
@davecosec, looks great besides the mandatory getDocumentId in the raw projection. We can add it later as an optional, once we discuss if we'd like to give some reading of the data here, for now, let's keep it simple.

oskardudycz (approved):
LGTM! 🤜🤛

PR #230, opened 2025-03-29
Added possibility to handle sequence of command handlers for a single stream as a single append

oskardudycz:


tburny:
It would be awesome if this feature is documented somewhere :) What do you think?

oskardudycz:
@tburny, indeed, it'll need to be documented :+1:

PR #235, opened 2025-04-19
Converted emmett-sqlite project to use Vitest

davecosec:


oskardudycz (changes requested):
@davecosec, sweet work, thank you. I think that what you did is better than replacing npm test fully, so allowing a gradual migration by having the step for vitest.

I'm happy to merge it, after you remove `test.db-shm` and test.db-wal` leftovers and ensure the rules around them are added to `.gitignore` :slightly_smiling_face:

oskardudycz (approved):
Nvm my previous review comment, I'm blind and can't read git diff correctly, I see that they were removed :sweat_smile:

PR #247, opened 2025-05-10
Add capability to specify custom checkpoint resolution by consumer or processor

oskardudycz:
You should not always use global position for checkpointing. EventStoreDB forces stream subscription to provide the stream position instead of the global one.

This adds the possibility to pass the getCheckpoint function to checkpointer as a way to tell that they should be using a different checkpointing strategy.

This fixes a bug on restarting the ESDB stream subscriber.


oskardudycz:
@mbirkegaard FYI

PR #250, opened 2025-05-22
[MongoDBEventStore] Fix bug in `projections.inline.count` to ignore any `null` projections

alex-laycalvert:
## Description

- Previously, the `projections.inline.count` helper attempted to exclude null projections by checking for `{ $exists: false }`
- This does not work for the MongoDB query where projections are explicitly set to `null`
- The new behavior ignores any non-existent or null projections
- Adjusted projection helper tests to account for new behavior


oskardudycz (approved):
Good catch @alex-laycalvert, and thank you for fixing it 👍

PR #258, opened 2025-07-31
feat: MongoDB Consumer based on MongoDB ChangeStream subscription

arturwojnar:


src/packages/emmett-mongodb/src/eventStore/mongoDBEventStore.subscription.e2e.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    consumer.reactor<ProductItemAdded>({
+      processorId: v4(),
+      eachMessage: (event) => {
+        assertTrue(receivedMessageCount <= 2);
+        assertEqual(
+          expectedProductItemIds[receivedMessageCount],
+          event.data.productItem.productId,
+        );
+
+        if (event.data.productItem.productId === lastProductItemId) {
```
oskardudycz, line 124: **SUGGESTION:** You can use a built-in stop after capabilities here to make this check easier. See https://github.com/event-driven-io/emmett/blob/bd65823671185d2aa422ffa17d30ae92b5c171ff/src/packages/emmett-esdb/src/eventStore/consumers/eventStoreDBEventStoreConsumer.handling.int.spec.ts#L189

src/packages/emmett-mongodb/src/eventStore/consumers/storeProcessorCheckpoint.ts
```diff
@@ -0,0 +1,69 @@
...
+      newPosition: Position;
+    }
+  | { success: false; reason: 'IGNORED' | 'MISMATCH' };
+
+export const storeProcessorCheckpoint = async <Position extends string | null>(
+  client: MongoClient,
+  options: {
+    processorId: string;
+    version: number | undefined;
+    newPosition: null extends Position
```
oskardudycz, line 22: **FIX:** In the end we need to support here also other types of positions. We'll need to be passing checkpoint as we're doing in ESDB. Without it, we'll be able to only use mongo processor in the MongoDB event store, and we want in the end to use it also with other event stores.

alex-laycalvert (commented):
I think this is a good start based off of the other consumers. Nothing that @oskardudycz hasn't already mentioned. I think some of the boillerplate can be removed since the addition of centralized consumer types and stuff.

Probably want to add some documentation that this consumer implementation requires change streams be enabled and can't be used in single-instance environments

src/packages/emmett-mongodb/src/eventStore/consumers/mongoDBEventsConsumer.ts
```diff
@@ -0,0 +1,270 @@
...
+
+export type MongoDBConsumerOptions<
+  ConsumerEventType extends Message = Message,
+> = EventStoreDBEventStoreConsumerConfig<ConsumerEventType> &
+  (
+    | {
+        connectionString: string;
+        clientOptions?: MongoClientOptions;
+        client?: never;
+        onHandleStart?: (
```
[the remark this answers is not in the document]
arturwojnar: Yes. I needed tha ton some point of the development, but now I dont :) Removed.

src/packages/emmett-mongodb/src/eventStore/consumers/mongoDBEventsConsumer.ts
```diff
@@ -0,0 +1,270 @@
...
+  >;
+  let isRunning = false;
+  const client =
+    'client' in options && options.client
+      ? options.client
+      : new MongoClient(options.connectionString, options.clientOptions);
+  const processors = options.processors ?? [];
+  const subscribe = _subscribe(
+    options.changeStreamFullDocumentPolicy,
+    client.db(),
```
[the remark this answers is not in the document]
arturwojnar: image
That happens here, right?

src/packages/emmett-mongodb/src/eventStore/consumers/subscriptions/index.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  return {
+    major: Number(versions?.[1]) || void 0,
+    minor: Number(versions?.[2]) || void 0,
+    bugfix: Number(versions?.[3]) || void 0,
+    rc: versions?.[4] || void 0,
+  };
+};
+
+export const generateVersionPolicies = async (db: Db) => {
```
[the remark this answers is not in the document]
arturwojnar: It checks the MongoDB version and returns a version-dependent configuration. Hm, actually, there is a difference between versions 5 and 6. What is your minimum support version?

src/packages/emmett-mongodb/src/eventStore/consumers/storeProcessorCheckpoint.ts
arturwojnar: esdb supports only the bigint
```typescript
export const storeProcessorCheckpoint = async <Position extends bigint | null>
```

Check out my changes, is it sth you were thinking of?

src/packages/emmett-mongodb/src/eventStore/consumers/mongoDBEventsConsumer.ts
```diff
@@ -0,0 +1,342 @@
...
+    },
+    stop: async () => {
+      if (stream) {
+        await stream.close();
+        isRunning = false;
+        runningPromise.resolve(null);
+      }
+    },
+    close: async () => {
+      if (stream) {
```
arturwojnar, line 313: the same as stop... :/

src/packages/emmett-mongodb/src/eventStore/consumers/mongoDBEventsConsumer.ts
```diff
@@ -0,0 +1,342 @@
...
+
+        const positions = await Promise.all(
+          processors.map((o) => o.start({ client } as Partial<HandlerContext>)),
+        );
+        const startFrom =
+          zipMongoDBMessageBatchPullerStartFrom<CheckpointType>(positions);
+
+        stream = subscribe<Event, CheckpointType>(startFrom);
+
+        void (async () => {
```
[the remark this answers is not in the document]
oskardudycz: @arturwojnar check ESDB implementation, there it’s handled with custom transform and callbacks https://github.com/event-driven-io/emmett/blob/bd65823671185d2aa422ffa17d30ae92b5c171ff/src/packages/emmett-esdb/src/eventStore/consumers/subscriptions/index.ts#L135

This gives a guarantee of sequential processing and passing it to producer. That should eliminate getting out of order and race conditions. I’m considering to also apply the similar pattern inside processors.

src/packages/emmett-mongodb/src/eventStore/consumers/storeProcessorCheckpoint.ts
oskardudycz: Yup, makes sense, now, thank you :+1:

src/packages/emmett-mongodb/src/eventStore/consumers/mongoDBEventsConsumer.ts
```diff
@@ -0,0 +1,282 @@
...
+          getFullDocumentValue: policy,
+          startFrom,
+        });
+      })();
+
+      return start;
+    },
+    stop: async () => {
+      if (stream.isRunning) {
+        await stream.stop();
```
arturwojnar, line 247: it's not the case for the esdb..

arturwojnar: @oskardudycz

oskardudycz (approved):
@arturwojnar, I added some additional fixes around starting/stopping/restarting, in general resilience alignment with other consumers.

I'll also need to check why some compatibility tests are not passing (currently commented out).

I also changed nested Mongo Resume Token to URN-based string formatted text, and added message position to it (as there can be more than one event appended in one changeset). I will generalise that to assume that checkpoint can be number, bigint or string, but in a separate PR.

Nevertheless, great work, thank you for doing it, and I appreciate your patience waiting for my final review. 👏 🥳

PR #264, opened 2025-10-01
Pass handleOptions to "read" which is consumed by readStream

wataruoguchi:
My understanding is that the "handleOptions" is consumed for getting the value of "partition" elsewhere. However, aggregateStream does not respect it. To maintain consistency, I propose passing the handleOptions to this "read" property.


wataruoguchi:
Wrong code. Please disregard. The right one is coming up soon

PR #336, opened 2026-05-06
Fixed race conditions in PostgreSQL event store consumer message polling 

oskardudycz:
I had some brain fog/fart and somehow didn't include (or probably removed somehow) the full transaction id handling I described in [How Postgres sequences issues can impact your messaging guarantees](https://event-driven.io/en/ordering_in_postgres_outbox/). In high contention, this could cause some messages to be lost in processors.

This PR restores proper filtering.

I also aligned PostgreSQL checkpointing to be a compound checkpoint built from the transaction id and global position.

I plan to add later, as a follow-up, some helpers to handle cases properly after restoring PostgreSQL from a snapshot or pg_dumb backup.

**BREAKING:** Made global position be string, as in some cases (like PostgreSQL, MongoDB) they're built from multiple values, they're not always bigint, but they will be always monotonic alphanumeric.

Thanks go to @nordfjord for catching this up.


src/packages/emmett-postgresql/src/eventStore/projections/postgresProjectionSpec.ts
```diff
@@ -121,9 +121,15 @@ export const PostgreSQLProjectionSpec = {
                 ...givenEvents,
                 ...Array.from({ length: numberOfTimes }).flatMap(() => events),
               ]) {
+                const checkpoint =
+                  PostgreSQLEventStoreCheckpoint.toProcessorCheckpoint({
+                    transactionId: ++globalPosition,
+                    globalPosition,
```
nordfjord, line 127: this looks like it might be a mistake? transactionId=globalPosition+1, globalPosition=globalPosition+1

I'd just put the +1's explicitly if that's what's intended

nordfjord: oh, I'm commenting on a test file, ignore me 🫡

src/packages/emmett-postgresql/README.md
```diff
@@ -106,7 +106,7 @@ const result = await eventStore.appendToStream<ShoppingCartEvent>(
 );
 
 console.log(result.nextExpectedStreamVersion); // 1n
-console.log(result.lastEventGlobalPosition); // Global position for ordering
+console.log(result.); // Global position for ordering
```
nordfjord, line 109: this is definitely a whoopsie

PR #355, opened 2026-07-09
Now 'END' means that processor start from last message, then continues from current checkpoint

oskardudycz:
Previously it was not possible to start from the end of the log, and then continue where processor left. 'CURRENT' position was confusing as it actually meant "start from the current position or beginning if you haven't processed anything". This PR cleans this and makes `CURRENT` obsolete (it'll be removed in the future releases), and make it clear where the initial starting point should be.

If someone wants to have the old behaviour, they can pass:

```{ts
 startFrom: 'END',
 checkpoints: 'DISABLED'
 }
```  

Or also 

```{ts
 startFrom: 'BEGINNING',
 checkpoints: 'DISABLED'
 }
```  

That'll make a volatile run without storing checkpoints anywhere and always starting from END or BEGINNING.

**Note:** ESDB and MongoDB handling will be fully aligned in a dedicated follow-up PR.


oskardudycz:
It's available from version `0.43.0-beta.25`

PR #357, opened 2026-07-12
Improve ESLint/Prettier related configs

onlyhans-dev:
- Removed redundant language specific overrides
- Disabled `editor.formatOnSave` for `.ts` files (it's handled by `source.fixAll.eslint` and `eslint-plugin-prettier`)
- Removed `.ts` files from Prettier RegEx
- Added `--cache` flag for both Prettier and ESLint scripts in `package.json`


oskardudycz:
@onlyhans-dev, thanks for doing that, makes total sense, and I'm curious why I didn't think about enabling cache earlier :facepalm: 

I added a commit fixing Prettier in samples.

I'll need to replicate that to [Pongo](https://github.com/event-driven-io/Pongo) 😀

PR #367, opened 2026-07-20
Added message handling middlewares to Command, Decider and Workflow handlers

oskardudycz:
Ok, so @MateuszNaKodach motivated me with his [question on LinkedIn](https://www.linkedin.com/feed/update/urn:li:activity:7483560145575153664?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A7483560145575153664%2C7484391306232864769%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287484391306232864769%2Curn%3Ali%3Aactivity%3A7483560145575153664%29) about how to map events to the HTTP statuses result.

It was possible even without any changes by doing one of (or combination of):
- checking the outcome of business logic, and throwing erros,
- accepting that each event is appended, so we record all "failure" decisions and map it,
- doing custom handling on event store methods.

Yet, none of that was accessible at the level I'd like, so I added options to specify the middleware that allows skipping, stopping processing or rejecting the whole batch.

Following the Boy Scout rule, I also added "before all" and "after all" processing. Bon apetit!

### DIFF
--- a/src/docs/api-reference/commandhandler.md
+++ b/src/docs/api-reference/commandhandler.md
@@ -19,16 +19,17 @@ It builds on the event store's [`aggregateStream`](/api-reference/eventstore#agg
 
 ### CommandHandlerOptions
 
-| Property               | Type                                          | Description                                                                        |
-| ---------------------- | --------------------------------------------- | ---------------------------------------------------------------------------------- |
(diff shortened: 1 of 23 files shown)


oskardudycz:
Available in `0.43.0-beta.31`

PR #368, opened 2026-07-20
Enhance Command and Decider handlers to support StoredEvent type for schema versioning

thiagomini:
There was no way to specify the `StoredEvent` type when using a `DeciderCommandHandler`, so when they differed,  the `upcast` function would issue a type error. Example:

```ts
type ProductItemAdded   = Event<'ProductItemAdded', { productId: string; quantity: number }>;
type ProductItemAddedV1 = Event<'ProductItemAdded', { productId: string; quantity: string }>;

DeciderCommandHandler<Cart, AddProductItem, ProductItemAdded>({
  decide, evolve, initialState,
  schema: {
    versioning: {
      upcast: (event: ProductItemAddedV1) => ({ /* ... */ }), // This would error - `upcast` function thinks the `event` type = `ProductItemAdded`
    },
  },
});
```

Now we can provide the `StoredType` as the fourth parameter:

```ts
DeciderCommandHandler<Cart, AddProductItem, ProductItemAdded, ProductItemAddedV1>({
  decide, evolve, initialState,
  schema: {
    versioning: {
      // event is inferred as ProductItemAddedV1 — no cast needed
      upcast: (event) => ({
        type: 'ProductItemAdded',
        data: {
          productId: event.data.productId,
          quantity: Number(event.data.quantity),
        },
      }),
    },
  },
});
```

### DIFF
--- a/src/packages/emmett/src/commandHandling/handleCommand.ts
+++ b/src/packages/emmett/src/commandHandling/handleCommand.ts
@@ -121,8 +121,12 @@ export type CommandHandlerFunction<State, StreamEvent extends Event> = (
 ) => StreamEvent | StreamEvent[] | Promise<StreamEvent | StreamEvent[]>;
 
 export const CommandHandler =
-  <State, StreamEvent extends Event, StoredEvent extends Event = StreamEvent>(
-    options: CommandHandlerOptions<State, StreamEvent, StoredEvent>,
+  <
+    State,
+    StreamEvent extends Event,
+    EventPayloadType extends Event = StreamEvent,
+  >(
+    options: CommandHandlerOptions<State, StreamEvent, EventPayloadType>,
   ) =>
   async <Store extends EventStore>(
     store: Store,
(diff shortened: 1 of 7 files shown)


src/packages/emmett/src/commandHandling/handleCommand.ts
```diff
@@ -384,7 +380,7 @@ const toAppendOptions = <
       ? {
           schema: options.schema as AppendToStreamOptions<
             StreamEvent,
-            EventPayloadType
+            StoredEvent
```
thiagomini: Edit: Thinking better about this, I'll revert, EventPayloadType makes more sense in this specific context.

This was renamed to be consistent with the type name used in the existing `DeciderCommandHandlerOptions`

oskardudycz (approved):
@thiagomini looks good to me! I also added 
- a small name alignment,
- fix for threading schema options in the event store,
- versioning tests for the regular handler.

PR #372, opened 2026-07-27
Add postgresql example for honojs integration

dawidhermann:
Adds sample project honojs emmett integration and postgres


oskardudycz:
Sweet! Thanks a lot, @dawidhermann! Much appreciated :)

I will need to simplify this observability setup, but I'll align it when it's time.

@kaluzaaa FYI

PR #373, opened 2026-07-29
Consumer unification

oskardudycz:
Consumers have become more mature and similar, so now it's time to unify processing across them to reduce the burden.

This PR extracts the shared abstraction for:
- consumer (previously we had type, now we also have implementations),
- message source (with pull- and push-based, plus in-memory),

Besides that, I also fixed integration test parallelism so they are stable and faster to run locally.


oskardudycz:
@dilgerma FYI

PR #378, opened 2026-08-07
Add missing emt_messages index for the consumer poll query

Bazze:
## Problem

`emt_messages` has no index beyond its primary key `(stream_id, stream_position, partition, is_archived)`, so the consumer poll in `readMessagesBatch` sequentially scans and top-N sorts the whole partition on every tick — including the caught-up poll that returns zero rows. Cost is O(table size) and independent of result size.

This is a permanent standing load rather than a slow-query-under-load problem. An idle read returns nothing, takes the backoff branch, and pins at the hardcoded 1000 ms ceiling (`pollingMessageSource.ts`), so `pullingFrequencyInMs` is unreachable at idle. Every consumer group therefore full-scans the table roughly once per second, forever, regardless of configuration.

On one production Aurora deployment this single statement accounted for **91% of total database load at only ~100k rows** (RDS Performance Insights, `db.load.avg` grouped by `db.sql_tokenized`). `ReadIOPS` was 0 and buffer cache hit ratio 100%, so it never surfaced as an IO problem — it is pure in-memory CPU burn.

## Change

Adds one index: `(transaction_id, global_position)`. It matches the poll's cursor and `ORDER BY` exactly (`(transaction_id, global_position) > ($tx, $gp) ORDER BY transaction_id, global_position`), so the `LIMIT` stops the scan instead of sorting. It also serves `readLastCommittedMessageCheckpoint`, whose reverse `LIMIT 1` has no cursor to bound it and today sorts the whole partition on every call.

Measured by `EXPLAIN (ANALYZE, BUFFERS)` at 20k rows across four cursor positions (caught up, 100 behind, half behind, replay from zero): **2–5 buffers with the index, 312 buffers plus a Sort without it**, in every regime.

Only one index, deliberately. A second index on `(global_position)` alone was measured and is **never chosen** on this query shape — a row comparison can only seek an index led by `transaction_id`. With that index alone the poll reverts to a sequential scan. It would cost write amplification on the hottest table for no read benefit. (It is warranted on 0.42.x, where the cursor is still a plain `global_position >= …` predicate; see the companion backport.)

`partition` and `is_archived` are omitted on purpose. `emt_messages` is `PARTITION BY LIST (partition)` then `LIST (is_archived)`, and indexes on a partitioned parent are created per leaf, so both columns are constant within every leaf the index lands on.

## Why a post-schema migration rather than `messagesTableSQL`

The DDL is wired into `eventStoreSchemaMigrations` *after* `schemaMigration`, not added to `messagesTableSQL`. `schemaSQL` is hashed into the long-shipped `emt:postgresql:eventstore:initial` migration, and dumbo aborts with `Migration hash mismatch` on any change, rolling back the whole batch — so every existing database would fail to start. Passing `ignoreMigrationHashMismatch` is not a workaround either: it updates the stored hash and returns *without executing the DDL*, so those databases would start but never get the index. Both behaviours were reproduced against a live database before choosing this shape.

Running after `schemaMigration` covers fresh and existing databases through one path and leaves the shipped hash untouched. The migration is deliberately kept out of `migrations_0_44_0`, since that array is destined for `pastEventStoreSchemaMigrations`, which runs before the tables exist.

The migration also drops `idx_messages_global_position` if present, to clean up databases coming from the 0.42.3 backport.

## Operational note

`CREATE INDEX` on a partitioned table takes a `ShareLock` on the parent and every leaf, blocking writes while it builds. `CREATE INDEX CONCURRENTLY` cannot replace it: PostgreSQL rejects it outright on a partitioned table, and it cannot run inside the migration transaction. On a large existing event store this migration is a write stall worth scheduling. The out-of-band escape hatch, if wanted later, is `CREATE INDEX ON ONLY <parent>` per level, `CREATE INDEX CONCURRENTLY` per leaf, then `ALTER INDEX … ATTACH PARTITION`.

## Tests

`messagesPollIndex.int.spec.ts` asserts the index by name against `EXPLAIN` output for each cursor regime, requires no `Sort` node, and scopes the negative to the specific leaf relation. It also covers inheritance by partitions created after the fact via `emt_add_partition`, and the reverse scan for `readLastCommittedMessageCheckpoint`. `migrations/current/migration.int.spec.ts` gains coverage for upgrading a database that already recorded the `initial` migration, and for dropping the backport's index.

The fixture is sized past the point where the planner still picks a bitmap scan plus a Sort, so the assertions reflect what a real event store does rather than a small-table artifact. `readMessagesBatchSQL` was extracted from `readMessagesBatch` so the tests `EXPLAIN` the real query rather than a copy that could drift; behaviour is otherwise unchanged.

## Not addressed here

Deliberately left alone, happy to open separate issues: no `lock_timeout` on the write-blocking migration; the duplicate `schemaSQL`/`schemaMigration` definitions in `schema/index.ts` and `schema/migrations/index.ts`; `migrations_0_44_0` still being unwired; and `postgreSQLMessageSource.readBatch` calling `readMessagesBatch` without a `partition`, so multi-partition consumers always poll the default.

### DIFF
--- a/src/packages/emmett-postgresql/src/eventStore/schema/messagesPollIndex.int.spec.ts
+++ b/src/packages/emmett-postgresql/src/eventStore/schema/messagesPollIndex.int.spec.ts
@@ -1,0 +1,203 @@
+import { dumbo, SQL } from '@event-driven-io/dumbo';
+import { pgDumboDriver, type PgPool } from '@event-driven-io/dumbo/pg';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import {
+  sharedPostgreSQLDatabase,
+  type PostgreSQLTestDatabase,
+} from '../../testing/postgreSQLTestDatabase';
+import { createEventStoreSchema } from '.';
+import {
+  readMessagesBatchSQL,
+  type PostgreSQLEventStoreCheckpoint,
+} from './readMessagesBatch';
+import { defaultTag } from './typing';
+
+// emt_add_partition sanitizes 'emt:default' into this leaf name.
+const defaultActiveLeaf = 'emt_messages_emt_default_active';
+const pollIndexName = 'idx_messages_transaction_id_global_position';
+
+// Below roughly this size the planner picks a bitmap scan plus a Sort, so the
+// no-Sort assertions would fail on a table too small to represent a real event store.
+const seededTransactions = 10;
+const messagesPerTransaction = 2000;
+const seededMessages = seededTransactions * messagesPerTransaction;
+
+void describe('emt_messages consumer poll index', () => {
+  // Needs its own database, not its own container: ANALYZE is not transactional for
+  // reltuples/relpages, so this seed's statistics would otherwise outlive the spec.
+  let database: PostgreSQLTestDatabase;
+  let pool: PgPool;
+
+  beforeAll(async () => {
+    database = await sharedPostgreSQLDatabase();
+    const connectionString = database.connectionString;
+    pool = dumbo({
+      connectionString,
+      driver: pgDumboDriver,
+      transactionOptions: { allowNestedTransactions: true },
+      pooled: false,
+    });
+
+    await createEventStoreSchema(connectionString, pool);
+    await seedMessages();
+  });
+
+  afterAll(async () => {
+    await pool?.close();
+    await database?.close();
+  });
+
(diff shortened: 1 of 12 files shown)


oskardudycz (approved):
@Bazze, looks good to me. I moved the PostgreSQL indexes migration to 0.43.0 and added it to the schema SQL, as it should be applied from the get-go. The migration hash issue will be tackled separately.

PR #379, opened 2026-08-07
Backport to 0.42.3: add missing `emt_messages` indexes for the consumer poll query

Bazze:
## Problem

`emt_messages` has no index beyond its primary key, so the consumer poll in `readMessagesBatch` sequentially scans and top-N sorts the whole partition on every tick — including the caught-up poll that returns zero rows. Cost is O(table size) and independent of result size.

This is a permanent standing load, not a slow-query-under-load problem. An idle read returns nothing and pins at the hardcoded 1000 ms backoff ceiling, so `pullingFrequencyInMs` is unreachable at idle and every consumer group full-scans the table roughly once per second, forever.

On one production Aurora deployment this single statement accounted for **91% of total database load at only ~100k rows**. `ReadIOPS` was 0 and buffer cache hit ratio 100%, so it never surfaced as an IO problem — it is pure in-memory CPU burn.

## Change

Adds two indexes, because on this version the poll filters on `global_position` directly (`AND global_position >= <cursor>`) while ordering by `transaction_id, global_position`, and neither index covers both cases:

- `(global_position)` serves the caught-up and slightly-behind polls, where the cursor is the only selective term — `transaction_id < xmin` matches nearly every row.
- `(transaction_id, global_position)` matches the `ORDER BY`, so a backlogged consumer stops at the `LIMIT` instead of sorting everything past the cursor. Without it, draining a backlog is quadratic in backlog size.

Verified by `EXPLAIN` at 20k rows: the caught-up poll picks the single-column index, replay-from-zero picks the composite, and no cursor position falls back to a partition scan. Shipping only the composite would leave the dominant caught-up poll walking the index from the low end, which grows with table size.

`partition` and `is_archived` are omitted on purpose. `emt_messages` is `PARTITION BY LIST (partition)` then `LIST (is_archived)`, and indexes on a partitioned parent are created per leaf, so both columns are constant within every leaf.

## Why a post-schema migration rather than `messagesTableSQL`

The DDL is wired into `eventStoreSchemaMigrations` *after* `schemaMigration`, not added to `messagesTableSQL`. `schemaSQL` is hashed into the long-shipped `emt:postgresql:eventstore:initial` migration, and dumbo aborts with `Migration hash mismatch` on any change, so every existing database would fail to start. Passing `ignoreMigrationHashMismatch` is not a workaround: it updates the stored hash and returns without executing the DDL.

## Upgrading to 0.43

0.43 rewrites the cursor as a row comparison on `(transaction_id, global_position)`. That makes the composite sufficient and leaves `(global_position)` unreachable, since a row comparison can only seek an index led by `transaction_id`. The 0.43 migration therefore creates the composite if absent and drops `idx_messages_global_position`, so an upgraded database ends up matching a fresh 0.43 install instead of carrying an index maintained on every append and never read.

The two migrations carry distinct names (`…:0.42.3:add-messages-poll-indexes` and `…:0.44.0:add-messages-poll-index`), so the upgrade applies both in order. Reusing one name with different SQL would trip dumbo's name + hash check and abort the upgrade.

## Operational note

`CREATE INDEX` on a partitioned table takes a `ShareLock` on the parent and every leaf, blocking writes while the indexes build. `CREATE INDEX CONCURRENTLY` cannot replace it: PostgreSQL rejects it outright on a partitioned table, and it cannot run inside the migration transaction. On a large existing event store this migration is a write stall worth scheduling.

## Tests

`messagesPollIndexes.int.spec.ts` asserts both indexes exist on the default active leaf, that no cursor position falls back to a partition scan, and — the part that justifies shipping two — that the caught-up poll uses the single-column index while replay-from-zero uses the composite. The existing migration and schema-creation suites pass unchanged.

### DIFF
--- a/src/packages/emmett-postgresql/src/eventStore/schema/index.ts
+++ b/src/packages/emmett-postgresql/src/eventStore/schema/index.ts
@@ -11,13 +11,13 @@ import {
 import type { PostgresEventStoreOptions } from '../postgreSQLEventStore';
 import { type PostgreSQLProjectionHandlerContext } from '../projections';
 import { appendToStreamSQL } from './appendToStream';
-import { migration_0_42_3_addMessagesPollIndexes } from './migrations/0_42_3';
 import { migration_0_38_7_and_older } from './migrations/0_38_7';
 import {
   migration_0_42_0_2_AddProcessorProjectionFunctions,
   migration_0_42_0_3_FixProcessorLockTimeout,
   migration_0_42_0_FromSubscriptionsToProcessors,
 } from './migrations/0_42_0';
+import { migration_0_42_4_addMessagesPollIndexes } from './migrations/0_42_4';
 import {
   releaseProcessorLockSQL,
   tryAcquireProcessorLockSQL,
@@ -84,7 +84,7 @@ export const eventStoreSchemaMigrations: SQLMigration[] = [
   // After schemaMigration, since it indexes tables schemaMigration creates. Adding the
   // DDL to messagesTableSQL instead would change schemaSQL, which is hashed into the
   // shipped 'initial' migration and would abort every existing database on mismatch.
-  migration_0_42_3_addMessagesPollIndexes,
+  migration_0_42_4_addMessagesPollIndexes,
 ];
 
 export type CreateEventStoreSchemaOptions = {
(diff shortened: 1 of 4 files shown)


oskardudycz (approved):
@Bazze, thanks for taking care of that, I moved the migration to `0.42.4`, as this will be the version under which it'll be released :+1:

Bazze:
Amazing, thanks! I applied the indexes myself yesterday via the hook we created into the emmett migration setup, and it brought down our DB load to almost nothing. So it was indeed the right fix 😊 we named the indexes the same as in this PR, so hopefully we should be good to bump versions without any clashes

oskardudycz:
@Bazze, thanks for catching and fixing it and checking in the real system. One of the reasons why I was postponing adding more indexes (for too long) was to have a real verification and test-first approach to confirm it; that's also why I appreciate those explain plans and tests.

Issue #328, opened 2026-04-16
pg_try_advisory_xact_lock in processor locks breaks with AWS RDS Proxy


Bazze:
## Problem

Emmett's processor lock functions (`emt_try_acquire_processor_lock`) use `pg_try_advisory_xact_lock`. AWS RDS Proxy [does not pin connections for transaction-scoped advisory locks](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-pinning.html), so the lock provides no cross-replica exclusivity. Multiple replicas acquire the lock simultaneously and process the same events.

## Idea for resolution

Replace `pg_try_advisory_xact_lock(p_lock_key)` with `pg_try_advisory_lock(p_lock_key)` in the processor lock acquisition functions. Session-scoped locks cause RDS Proxy to pin the connection.

It might be necessary to investigate and see how locks are released since it goes from transaction to session based.

Bazze:
We hit this in production on AWS RDS Proxy and worked around it out-of-tree, so I dug through the lock code to see whether the plug-and-play lock interface you @oskardudycz  floated has a clean seam. It think it does, and it's smaller than I expected. Line refs below are against `0.43.0-beta.38`.

## The seam is one function

Everything lock-related funnels through a single decorator: `wrapHooksWithProcessorLocks` ([emmett-postgresql/src/eventStore/consumers/postgreSQLProcessor.ts:344](https://github.com/event-driven-io/emmett/blob/main/src/packages/emmett-postgresql/src/eventStore/consumers/postgreSQLProcessor.ts#L344)), which wraps `ProcessorHooks.onStart`/`onClose` and is called from all three PG processor factories (projector `:384`, reactor `:495`, workflow `:560`).

Core [processors.ts](https://github.com/event-driven-io/emmett/blob/main/src/packages/emmett/src/processors/processors.ts) has no lock concept at all — it runs `hooks.onStart(context)` (`:727`) and then `checkpointer.read()` (`:731`). That's already the correct ordering for a lock, so hoisting the decorator into core is most of the work.

Three things make it easier than it looks:

[...]

PR #382, opened 2026-08-12
Hoist processor locking into a core `ProcessorLock` interface

Bazze:
## Why

Locking lives in a PostgreSQL-specific hook decorator (`wrapHooksWithProcessorLocks`), so ESDB, MongoDB and SQLite processors have no way to lock at all, and no external service can supply one. This is the first (of probably three PRs) discussed in #328 — it moves the seam into core and adapts PostgreSQL to it, without adding any new locking capability.

## What changed

The processor now owns the lock lifecycle. It acquires inside the processing scope before reading the checkpoint, applies the acquisition policy, and releases on close ahead of the close hooks, which may tear down the connection the lock is held on.

`ProcessorLock` and `LockAcquisitionPolicy` are new in `@event-driven-io/emmett`, and `BaseMessageProcessorOptions` gained `lock?: { lock?, acquisitionPolicy? }`. `LockAcquisitionPolicy` is re-exported from `@event-driven-io/emmett-postgresql`, so existing imports keep working. ESDB, MongoDB and SQLite inherit the option through their existing option spreads, with no code change.

`postgreSQLProcessorLock` keeps its signature and its SQL, and is now typed as `ProcessorLock<PostgreSQLProcessorLockContext>`. It no longer throws or retries by itself — that logic moved to core, where the `// TODO: This should be moved o prcessor` comment asked for it.

## Behaviour changes

Projectors acquire the lock before `truncateOnStart` runs, where the truncate used to come first. Under `rebuildPostgreSQLProjections`' default retry policy, a losing instance previously deleted every projection row and then held those row locks for the length of its retry loop. It now blocks on the lock first.

`postgreSQLProcessorLock`'s `lockAcquisitionPolicy` option is deprecated and ignored, since the policy is the processor's job now. No in-repo caller passes it for anything but `skip`, which was the no-throw case, but a direct caller outside the processors would see `tryAcquire` return `false` where it previously threw.

The PostgreSQL processor `lock` option gained an optional `lock`, so a caller can supply their own `ProcessorLock` instead of the message-store one. The intersection with the core option made it typecheck already; honouring it beats silently dropping it.

## Compatibility

No export was removed from either package and no exported type narrowed — verified by diffing the emitted `dist/index.d.ts` for both packages against `main`, and by compiling a consumer against them. `tryAcquireProcessorLockWithRetry` was deleted, but it was never in a barrel and the package has no wildcard subpath export, so it was never reachable.

## Testing

`postgreSQLProcessorLock.int.spec.ts` passes unmodified (26 cases) — it drives the SQL directly, which does not move, so it is the regression proof for this PR. New: a core unit spec for the policy and lifecycle, a store-agnostic conformance suite in `emmett-tests` that PR2 can reuse, and an int case pinning the projector's lock wiring. Full PostgreSQL int suite and the workspace unit suite are green.

## Not in this PR

Leases, `renew`/heartbeat and loss handling, and the Redis package. No `noopProcessorLock` either — absence of a lock means no lock.

## Open questions

Is the nested `lock: { lock, acquisitionPolicy }` shape right, or would top-level `lock?: ProcessorLock` plus `lockOptions?` read better? Nested keeps the existing PostgreSQL option shape unchanged, which is why it is proposed.

`timeoutSeconds` stays construction-time on the PostgreSQL lock and core never sees it, since it is meaningless for a lock without a lease. Happy to move it if you would rather core carry it.

On an exhausted `retry` policy, `asyncRetry` rejects from inside its own loop, so the error is `Retrying because of result: false` rather than the tailored `Failed to acquire lock for processor '<id>'`. That is pre-existing behaviour carried over verbatim; fixing it means touching `asyncRetry`, so it is left alone here.

### DIFF
--- a/src/packages/emmett-postgresql/src/eventStore/consumers/postgreSQLEventStoreConsumer.handling.int.spec.ts
+++ b/src/packages/emmett-postgresql/src/eventStore/consumers/postgreSQLEventStoreConsumer.handling.int.spec.ts
@@ -5,6 +5,8 @@ import {
   assertThatArray,
   assertThrowsAsync,
   defaultTag,
+  getProjectorId,
+  unknownTag,
   type Event,
 } from '@event-driven-io/emmett';
 import { v4 as uuid } from 'uuid';
@@ -1099,6 +1101,46 @@ void describe('PostgreSQL event store started consumer', () => {
           processorId,
           startFrom: 'CURRENT',
           eachMessage: () => {},
+        });
+
+        try {
+          await assertThrowsAsync<EmmettError>(
+            () => consumer.start(),
+            (error) => error.message.includes(processorId),
+          );
+        } finally {
+          await consumer.close();
+        }
+      },
+    );
+
+    void it(
+      'fails to start a projector when another instance holds the processor lock',
+      withDeadline,
+      async () => {
+        const projectionName = `lockedProjection-${uuid()}`;
+        const processorId = getProjectorId({ projectionName });
+
+        const otherInstanceLock = postgreSQLProcessorLock({
+          processorId,
+          version: 1,
+          partition: defaultTag,
+          processorInstanceId: 'other-crashed-instance',
+          projection: {
+            name: projectionName,
+            kind: unknownTag,
+            version: 1,
+            handlingType: 'async',
+          },
+        });
+        await pool.withConnection((connection) =>
+          otherInstanceLock.tryAcquire({ execute: connection.execute }),
+        );
+
+        const consumer = postgreSQLEventStoreConsumer({ connectionString });
+        consumer.projector<GuestStayEvent>({
+          projection: collectingProjection([], projectionName),
+          startFrom: 'CURRENT',
         });
 
         try {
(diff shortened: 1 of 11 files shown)


src/packages/emmett-postgresql/src/eventStore/consumers/postgreSQLEventStoreConsumer.handling.int.spec.ts
```diff
@@ -1112,6 +1114,46 @@ void describe('PostgreSQL event store started consumer', () => {
       },
     );
 
+    void it(
+      'fails to start a projector when another instance holds the processor lock',
```
oskardudycz: **CONCERN:** This claim isn't precise. It should only fail, if the lock is held consistently during the retry policy. The processor should attempt to retry the lock according to the retry policy. It might just be a test naming issue, but it's worth double-checking.

oskardudycz (commented):
Looks good to me, in general, but please comments before approval.

src/packages/emmett-postgresql/src/eventStore/consumers/postgreSQLEventStoreConsumer.handling.int.spec.ts
Bazze: Good catch, it's a naming issue. This test exercises the default fail policy (DefaultPostgreSQLProcessorLockPolicy, unchanged by this PR), so the claim is only true for that policy. I will rename it and the reactor test above it to say so. Retry is honoured and already covered by the test `consumer waits for lock timeout before taking over a crashed processor`, which retries while the lock is held and takes over on expiry. I verified it's a real check by swapping its policy to fail, which makes it throw on the first attempt.

oskardudycz: Makes sense, thank you :+1:

PR #383, opened 2026-08-13
Add Honojs documentation

oskardudycz:
Also enable setting up Emmett conventions based on the existing honojs setup.

The follow up to: https://github.com/event-driven-io/emmett/pull/380

### DIFF
--- a/src/docs/frameworks/honojs.md
+++ b/src/docs/frameworks/honojs.md
@@ -1,0 +1,294 @@
+---
+documentationType: reference
+outline: deep
+---
+
+# Hono Integration
+
+`@event-driven-io/emmett-honojs` provides Hono adapters for exposing commands and read models through a Web API. Use it when you already have a Hono API, prefer Hono for a new service, or want a Fetch-native HTTP boundary with a small integration layer.
+
+Routes remain ordinary Hono routes. A command route reads and validates the request, obtains any external data the decision needs, calls the command handler, and chooses a response. A query route reads its parameters, queries the read store, and returns the matching model. The integration supplies repeated HTTP boundary work around both kinds of route.
+
+Nothing in the application setup or response handling requires event sourcing. Use the command-result, ETag, and event-backed testing helpers only where those concerns are part of the API.
+
+## What the Integration Adds
+
+| Concern                | What the integration provides                                                             | What remains in the route or application                  |
+| ---------------------- | ----------------------------------------------------------------------------------------- | --------------------------------------------------------- |
+| Hono setup             | Feature-route registration, generated ETags, and Problem Details errors.                  | Authentication, tracing, and application-specific routes. |
+| HTTP responses         | Helpers for status, body, `Location`, ETag, and Problem Details responses.                | Choosing the response that represents each outcome.       |
+| Command results        | Mapping returned events to success or failure responses.                                  | Deciding whether business failures are returned or thrown |
+| Optimistic concurrency | Reading `If-Match` and returning a stream version as an ETag.                             | Choosing which writes require a client-supplied version.  |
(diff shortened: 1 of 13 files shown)


oskardudycz:
@dawidhermann, could you have a look to see if I didn't mess anything up? :sweat_smile:

src/packages/emmett-honojs/src/application.ts
```diff
@@ -17,40 +17,55 @@ export type ApplicationOptions = {
...
+  return application;
+};
+
+export const configureApplication = (
+  application: Hono,
+  options: ApplicationOptions,
+): Hono => {
+  const { apis, mapError, disableProblemDetailsMiddleware } = options;
+
+  application.use(etag());
```
dawidhermann, line 40: Could we avoid installing `etag()` globally, or restrict it to cacheable GET/HEAD responses?

Hono's ETag middleware runs after the handler and converts a response to 304 when its ETag matches `If-None-Match`. For an exact ETag match, this is not restricted to safe HTTP methods.

For example, a POST command returning an explicit stream-version ETag behaves as follows:

```ts
let executed = false;

router.post('/command', (context) => {
  executed = true;
  return NoContent({
    context,
    eTag: toWeakETag(4),
  });
});
```
A POST /command with If-None-Match: W/"4" executes the command (executed === true) but returns 304 instead of the command response.

The middleware also generates ETags for Problem Details responses, so a repeated failing request with a matching If-None-Match can turn a 500 into a 304.

Since this package already uses explicit ETags as stream versions, I think generated cache validators should either be opt-in or limited to successful, cacheable representations.

src/packages/emmett-honojs/src/application.ts
```diff
@@ -17,40 +17,55 @@ export type ApplicationOptions = {
...
   // add problem details middleware
   if (!disableProblemDetailsMiddleware) {
-    const errorMapper = mapError ?? defaultErrorToProblemDetailsMapping;
-    app.onError((error, c) => {
+    application.onError((error, c) => {
       // Replace with different mechanism as each app instance can only have defined one onError handler
-      const problemDetails = errorMapper(error);
-      return c.json(
+      const problemDetails =
+        mapError?.(error) ?? defaultErrorToProblemDetailsMapping(error);
```
dawidhermann, line 49: Just out of curiosity - does this preserve Hono's native `HTTPException` status if i throw
`throw new HTTPException(403, { message: 'Forbidden' });`?

dawidhermann:
To check - there might be some issues with generic types definitions related to child routers
https://github.com/orgs/honojs/discussions/2257

src/docs/frameworks/honojs.md
```diff
@@ -0,0 +1,294 @@
...
+
+Hono reads JSON, query string values, and route parameters through `context.req`. The integration does not install Express-style body parsers.
+
+### Add the Defaults to an Existing Hono API
+
+`configureApplication(application, options)` applies the same setup without creating the application. Existing infrastructure routes and middleware stay on the provided app. Here `/health` remains available beside the configured API stack:
+
+<<< @./../packages/emmett-honojs/src/e2e/configureExistingApplication.e2e.spec.ts#configure-existing-application
+
+`configureApplication` calls `application.onError` when Problem Details handling is enabled. Hono has one effective application error handler, so set `disableProblemDetailsMiddleware: true` and call `application.onError` yourself when the application already owns that contract:
```
dawidhermann, line 68: If you have subrouters they have their own `onError` which ignores the parent's one
Check this unit test:
https://github.com/honojs/hono/blob/41bdc4288ed7023c24d35f32bd065bd963435e3e/src/hono.test.ts#L1874-L1925

oskardudycz:
@dawidhermann 
image

All good on the calls; I'll handle that in a dedicated PR, since this PR grew too big, and I also need to align the Express package accordingly. I'll tag you in them.

PR #403, opened 2026-09-21
Added Cloudflare Durable Objects event store implementation

oskardudycz:
Ok, @EisenbergEffect this means that we have the first version of the Durable Objects event store 🚀

The consumer part works, but to make it work, I'll need to use the [Alarms API](https://developers.cloudflare.com/durable-objects/api/alarms/) to keep the Durable Object alive and wake it up to pull messages. Still, I'll need to rethink that part carefully. 

Samples will come soon after.

Still, sync api like appending, reading events, Pongo projections, etc. should work already 🙂

@SamHatoum FYI


oskardudycz:
Changes are released under `0.43.0-beta.48`, I'm working on porting Pongo samples :slightly_smiling_face:

